PR #8, opened 2019-09-11
Add `order` parameter to collections that support it

jhernand:
This patch adds the description of the `order` parameter to the
collections of the clusters management service that support it.


jhernand:
FYI: @tiwillia

PR #22, opened 2019-10-07
Add `cpu_total_by_node_roles_os` metric query

jhernand:


model/clusters_mgmt/v1/metric_queries_resource.model
```diff
@@ -0,0 +1,24 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages metric queries for a cluster.
+resource MetricQueries {
+	// Reference to the resource that retrieves 24 hour history of the amount of total cpu
```
cben: I don't think "24 hour history" is relevant here, probably copy-paste mistake.  Do we also say this in code or openapi in clusters-service?

jhernand: Yes, this is copied from the OpenAPI specification. If it isn't correct submit a pull request to fix it.

PR #38, opened 2019-10-24
Fix sku list

rawsyntax:
I ran into some problems using the sdk with all caps SKUS (I also tried SKUs). I suspect there is a problem somewhere in the plural snake case generation.

Here is an example snippet and the error generated:
```golang
        ctx := context.Background()
        token := os.Getenv("OCM_TOKEN")
	connection, err := sdk.NewConnectionBuilder().
		Tokens(token).
		BuildContext(ctx)
	if err != nil {
		glog.Fatalf("Can't build connection %v", err)
		os.Exit(1)
	}
	defer connection.Close()

        skusResource := connection.AccountsMgmt().V1().SKUS()

	skusResponse, err := skusResource.List().Send()
	if err != nil {
		glog.Fatalf("Can't list skus: %v", err)
	}
```

```
F1024 10:54:49.180540   81265 main.go:113] Can't list skus: expected kind 'SKU' or 'SKULink' but got 'Sku'
```

However naming as `Sku`, `Skus` etc works fine

### DIFF
--- a/model/accounts_mgmt/v1/root_resource.model
+++ b/model/accounts_mgmt/v1/root_resource.model
@@ -82,8 +82,8 @@ resource Root {
 	}
 
 	// Reference to the resource that manages the collection of
-	// SKUS
-	locator SKUS {
-		target SKUS
+	// Skus
+	locator Skus {
+		target Skus
 	}
 }
--- a/model/accounts_mgmt/v1/sku_resource.model
+++ b/model/accounts_mgmt/v1/sku_resource.model
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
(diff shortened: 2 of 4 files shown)


rawsyntax:
@jhernand please review

jhernand:
What is the server returning in the "kind" attribute of the SKU object? Should be "SKU", all upper case. Is it?

rawsyntax:
@jhernand thats the error -- it is returning `Sku` as the kind

jhernand:
@rawsyntax Can we change that in the server? I'd prefer that, if possible.

rawsyntax:
@jhernand ok I will try to fix it there

rawsyntax:
fixed in the server

PR #39, opened 2019-10-24
Allow search on accounts endpoint

rawsyntax:


rawsyntax:
@jhernand please review

PR #47, opened 2019-11-08
Add ids to role binding type

rawsyntax:
Account id in particular is required to create a role binding


rawsyntax:
@jhernand please review

PR #48, opened 2019-11-08
Release 0.0.19

jhernand:
The more relevant changes in the new version are the following:

- Added identifiers to role binding type.


jhernand:
FYI: @rawsyntax

PR #50, opened 2019-11-13
Query resource quota from root and delete by id

rawsyntax:
This enables deletion of resource quota via id and listing of resource quota without specifying an organization id


rawsyntax:
@jhernand please review

PR #51, opened 2019-11-14
Release 0.0.20

jhernand:
The more relevant changes in the new version are the following:

- Query resource quota from root and delete by identifier.


jhernand:
FYI: @rawsyntax

PR #58, opened 2019-11-19
Refactor flavour nodes

jhernand:
This patch changes the `Flavour` type so that it uses a separate type to
specify the number of nodes instead of the same type used by the
`Cluster` type.


jhernand:
@zgalor please review.

zgalor:
LGTM :+1:

PR #72, opened 2019-12-06
Use add-on installation as a wrapper around add-ons when installed in…

vkareh:
… a cluster

This change is needed in the SDK to fix build failures in https://gitlab.cee.redhat.com/service/uhc-clusters-service/merge_requests/1428

### DIFF
--- a/model/clusters_mgmt/v1/add_on_installation_type.model
+++ b/model/clusters_mgmt/v1/add_on_installation_type.model
@@ -1,0 +1,25 @@
+/*
+Copyright (c) 2019 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


model/clusters_mgmt/v1/add_on_installation_type.model
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of an add-on installation in a cluster.
+class AddOnInstallation {
+	// ID used to identify the cluster that this add-on is attached to.
+	ClusterID String
```
jhernand: This should be a link to the cluster: `link Cluster Cluster`.

model/clusters_mgmt/v1/add_on_installation_type.model
```diff
@@ -0,0 +1,25 @@
...
+
+	// Link to add-on attached to this cluster.
+	link Addon []AddOn
```
jhernand: This is a a link to a list of add-ons, but should be a link to a single add-on: `link Addon AddOn`.

model/clusters_mgmt/v1/add_on_installation_type.model
vkareh: Makes sense. Done

model/clusters_mgmt/v1/add_on_installation_type.model
vkareh: Good catch. Fixed.

PR #80, opened 2019-12-24
Fix add-on target in cluster resource

nimrodshn:
@cben @igoihman PTAL


igoihman:
LGTM

cben:
ah, some most of the work was already done in #72, just this remained.  LGTM.

PR #82, opened 2019-12-31
Add CreatedAt and UpdatedAt fields to the Subscription type

nimrodshn:
cc: @jhernand @tiwillia @igoihman @cben @Lir10 @tzvatot @gshilin-sdb

Current subscription model (and its derivatives, e.g. the openapi) does not contain the `CreatedAt` and `UpdatedAt` fields which **do exist** in the current API supplied by the service. These fields are needed for the `subscription-cleaner` CronJob we are planning the run.


Lir10:
@nimrodshn Looks good to us

tzvatot:
LGTM

PR #90, opened 2020-01-02
Add GCP flavour and change AWS flavour to contain also the instance type

tzvatot:
@jhernand @nimrodshn @zgalor @oourfali @igoihman PTAL

### DIFF
--- a/model/clusters_mgmt/v1/flavour_type.model
+++ b/model/clusters_mgmt/v1/flavour_type.model
@@ -18,14 +18,15 @@ limitations under the License.
 // with 10 infra nodes and 1000 compute nodes.
 class Flavour {
 	// Default _Amazon Web Services_ settings of the cluster.
-	//
-	// These can be overriden specifying in the clsuter itself a different set of settings.
 	AWS AWSFlavour
+
+	// Default _Google Cloud Platform_ settings of the cluster.
+	GCP GCPFlavour
 
 	// Number of nodes that will be used by default when creating a cluster that uses
 	// this flavour.
 	//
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/flavour_type.model
```diff
@@ -19,13 +19,18 @@ limitations under the License.
 class Flavour {
 	// Default _Amazon Web Services_ settings of the cluster.
 	//
-	// These can be overriden specifying in the clsuter itself a different set of settings.
+	// These can be overridden specifying in the cluster itself a different set of settings.
```
zgalor: not all can be overridden. only the compute instance type

model/clusters_mgmt/v1/flavour_type.model
```diff
@@ -19,13 +19,18 @@ limitations under the License.
...
+	// Default _Google Cloud Platform_ settings of the cluster.
+	//
+	// These can be overridden specifying in the cluster itself a different set of settings.
+	GCP GCPFlavour
+
 	// Number of nodes that will be used by default when creating a cluster that uses
 	// this flavour.
 	//
-	// These can be overriden specifying in the cluster itself a different number of nodes.
+	// These can be overridden specifying in the cluster itself a different number of nodes.
```
zgalor: Again, only the compute nodes can be overridden

tzvatot: Thanks. I'll remove this comment.

model/clusters_mgmt/v1/flavour_type.model
tzvatot: Fixed

PR #98, opened 2020-01-13
add missing list params of registry_credentials

petli-openshift:


petli-openshift:
I happened to find these params are missing in `ocm-sdk-go` when working on the `api-tests`. It's not blocking anything.

model/accounts_mgmt/v1/subscriptions_resource.model
```diff
@@ -59,6 +59,15 @@ resource Subscriptions {
...
 		in Order String
 
+		// Filter subscriptions by a comma separated list of labels:
+		//
+		// [source]
+		// ----
+		// env=staging,department=sales
+		// ----
+		//
+		in Labels String
```
jhernand: As the parameter already exists I will merge this change. But I think this should have been part of the `search` parameter:

```
labels.env = 'staging' and labels.department = 'sales'
```

Consider changing that.

petli-openshift: @jhernand I see your point. For now, we also support "dot" search for nested fields. For example, you can do `search="labels.key='env' and labels.value='staging'"`. However, `labels.env` is not correct since it's not a field of labels.

beside, for filtering multiple labels, we need a post step of `HAVING` clause, which is complex for end users. Since this is the only use case, we just created a separate param to make it simple.

PR #101, opened 2020-01-28
Add `Deleting` and `Removed` states

zgalor:
Add 2 new AWSInfrastructureAccessRoleGrantState possible states

### DIFF
--- a/model/clusters_mgmt/v1/aws_infrastructure_access_role_grant_state_type.model
+++ b/model/clusters_mgmt/v1/aws_infrastructure_access_role_grant_state_type.model
@@ -24,4 +24,12 @@ enum AWSInfrastructureAccessRoleGrantState {
 
 	// The attempt to grant access role to user ARN failed.
 	Failed
+
+	// The access role grant is in the process of being deleted.
+	Deleting
+
+	// This ia a special state intended for the user know
+	// that the access role grant has been removed by SRE.
+	// The user can delete this grant from the DB.
+	Removed
 }


zgalor:
@jhernand @tzvatot @nimrodshn @igoihman PTAL

tzvatot:
I suggest to drop the 'sre' from the comment.
Anyway LGTM

model/clusters_mgmt/v1/aws_infrastructure_access_role_grant_state_type.model
```diff
@@ -24,4 +24,10 @@ enum AWSInfrastructureAccessRoleGrantState {
 
 	// The attempt to grant access role to user ARN failed.
 	Failed
+
+	// The access role grant is being deleted.
+	Deleting
+
+	// The access role grant has been removed.
+	Removed
```
jhernand: It is a bit weird that one is `Deleting` and the other `Removed`. Should probably be `Deleting` and `Deleted` or `Removing` and `Removed`.

zgalor: I actually want them to be different.
`Deleting` is an action initiated by the user - at the end the grant record will not exists.
`Removed` is a state that the corresponding object in AWS has been removed by SRE, and the record is still in the DB to let the user know it has been removed.
The user will have to perform a delete action to actually delete the `removed` record from the DB

jhernand: OK, please explain that in the documentation of each value.

zgalor: Done

model/clusters_mgmt/v1/aws_infrastructure_access_role_grant_state_type.model
```diff
@@ -24,4 +24,12 @@ enum AWSInfrastructureAccessRoleGrantState {
 
 	// The attempt to grant access role to user ARN failed.
 	Failed
+
+	// The access role grant is in the process of being deleted.
+	Deleting
+
+	// This ia a special state intended for the user know
+	// that the access role grant has been removed by SRE.
+	// The user can delete this grant from the DB.
```
jhernand: So when the object is in this state the grant has actually been removed and has no effect but we will still be showing it in the API and asking the user to delete it? I don't understand what is the purpose of that. Please elaborate.

zgalor: So the idea is to let the user know SRE removed the access.
I don't want the user to think they created access, but then have it suddenly disappear.
That way the user knows what happend, and can contact SRE to get further information why their access has been removed

zgalor: It would be better if we had some general alert mechanism. Then we would just delete the grant entirely from the DB and raise an alert. But we don't, so considering current mechanisms, I think it's the best way to reconcile with aws account operator

jhernand: Why would SRE remove the access? I assume this would be done only if the user requests so, in which case there is no need to notify the user. In addition if we support removing the access in our API (not sure if we do) then it should never be removed directly but via our API.

PR #108, opened 2020-02-11
adding 'targetNamespace' and 'installMode' fields to addon_type model

zvikorn:
Signed-off-by: Tzvika Avni <tavni@dhcp-2-174.tlv.redhat.com>

### DIFF
--- a/model/clusters_mgmt/v1/add_on_type.model
+++ b/model/clusters_mgmt/v1/add_on_type.model
@@ -36,4 +36,10 @@ class AddOn {
 
 	// Used to determine how many units of quota an add-on consumes per resource name.
 	ResourceCost Float
+
+	// The namespace in which the addon CRD exists.
+	TargetNamespace String
+
+	// The mode in which the addon is deployed.
+	InstallMode AddOnInstallMode
 }
--- a/model/clusters_mgmt/v1/add_on_install_mode_type.model
+++ b/model/clusters_mgmt/v1/add_on_install_mode_type.model
@@ -1,0 +1,26 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


model/clusters_mgmt/v1/add_on_type.model
```diff
@@ -36,4 +36,10 @@ class AddOn {
 
 	// Used to determine how many units of quota an add-on consumes per resource name.
 	ResourceCost Float
+
+	// The namespace name in which the addon is installed.
+    TargetNamespace String
```
jhernand: Why is this needed?

model/clusters_mgmt/v1/add_on_type.model
```diff
@@ -36,4 +36,10 @@ class AddOn {
...
+
+    // The mode in which the addon is installed.
+    InstallMode String
```
jhernand: What are the possible values of this install mode and what do they mean? Consider adding a enumerated type for that:

```
enum AddOnInstallMode {
        // This mode means blah, blah, blah ...
        FirstMode

        // This mode means ....
        SecondMode
}
```

That should go in a separate `add_on_install_mode_type.model` file.

jhernand: Also please remember to indent the code only with tabs.

jhernand:
@vkareh please review.

model/clusters_mgmt/v1/add_on_type.model
zvikorn: In the documentation comments, you mean in the .model file itself?
This field is required sine we want to present the consumer with the information of which namespace the addon is installed in. We want to show this the targetNamespace on the addon tab.

jhernand: Yes, I mean in the documentation comments in the .model file.

model/clusters_mgmt/v1/add_on_type.model
vkareh: This should be an enum. Currently accepted values should be `AllNamespaces` and `SingleNamespace`.

zvikorn:
@vkareh could you please see if the description in each model is correct?

vkareh:
LGTM! thanks!

PR #109, opened 2020-02-13
Add State to AWSInfrastructureAccessRole

zgalor:


zgalor:
@jhernand @tzvatot @igoihman PTAL

Issue #111, opened 2020-02-24
openapi: listing collections don't document returning a `mind` field


cben:
I notice none of the `method List {` blocks contain an `out Kind String` parameter.  
And indeed the generated openapi documents `kind` for individual objects corresponding to `method Get` eg. `/api/clusters_mgmt/v1/clusters/{cluster_id}` but not for collection lists e.g. `/api/clusters_mgmt/v1/clusters`

@nimrodshn @jhernand  I can send a PR to add those, will that be enough?  

I see generated server code has special logic to return correct Kind, including link vs full object.
Will it automatically also return kind for lists?

jhernand:
Yes, the generated server code automatically returns the "kind" attribute also for lists. For example, this is the generated server code that returns the cluster list: https://github.com/openshift-online/ocm-sdk-go/blob/6997f79338819f63753b9156c9a8ecbb8537a230/clustersmgmt/v1/clusters_resource_json.go#L106-L157

And this is the part of the generator that takes care of that: https://github.com/openshift-online/ocm-api-metamodel/blob/a92c173afbff673a56219f427f20b20ab9a1778d/pkg/generators/golang/json_generator.go#L1007-L1041

I believe that what needs to be changed is the code that generates the OpenAPI specification. Most probably this method: https://github.com/openshift-online/ocm-api-metamodel/blob/a92c173afbff673a56219f427f20b20ab9a1778d/pkg/generators/openapi/openapi_generator.go#L316-L343

PR #112, opened 2020-02-24
README: more details on method blocks

cben:
I notice the meaning of parameters depends on method, so decided to document that better.

- see #111 about `Kind` parameter in List.

- wasn't sure where to document the `?method=Get` thing.  AFAICT `MethodMiddleware` code implementing this is specific to clusters-service, and SDK doesn't use it on client nor server side.  
  But it's used by some requests from the UI, so should be documented, and should appear in openapi (which adding to README doesn't achieve).

  An obvious place I thought is on root resource for clusters_mgmt, but its comment doesn't seem to get into the openapi.  The doc string for `/api/clusters_mgmt/v1` is "Retrieves the version metadata." which is hard-coded in the generator.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -117,9 +117,12 @@ method List {
 	// Order criteria.
 	in Order String
 
+        // Kind of the result - "ClusterList".
+        out Kind String
+
 	// Total number of items of the collection that match the search criteria,
 	// regardless of the size of the page.
-	in Total Integer
+	out Total Integer
 
 	// Retrieved list of clusters.
 	out Items []Cluster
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -117,20 +117,41 @@ method List {
 	// Order criteria.
 	in Order String
 
+        // Kind of the result - "ClusterList".
+        out Kind String
```
jhernand: This is not correct. The "kind" parameter is built-in and shouldn't appear in the specification.

README.adoc
```diff
@@ -117,20 +117,41 @@ method List {
...
 	// Retrieved list of clusters.
 	out Items []Cluster
 }
 ----
 
 Methods have _parameters_ defined by their direction (_in_ or _out_), their
-name and their type. In the above example there are five input parameters
-(named `Page`, `Size`, `Search` and `Order`) and one output parameter (named
+name and their type. In the above example there are four input parameters
+(named `Page`, `Size`, `Search` and `Order`) and four output parameter (named
```
jhernand: Should be: `two output parameters (named `Total` and `Items`).

README.adoc
```diff
@@ -117,20 +117,41 @@ method List {
...
 `Items`).
 
+- "List" conceptual method is implemented by http GET method, with
```
jhernand: `http` is an acronym, therefore it should be written using all upper case in the documentation.

README.adoc
cben: OK, dropping from here.  I'll document it's automatically included.

README.adoc
cben: oh, I forgot to edit this part.  I meant four, including the in-out Page and Size.

README.adoc
cben: done

PR #114, opened 2020-02-25
Fix indentation of README

jhernand:
In Asciidoc the way to have multiple paragraphs inside a bullet point is
to join them with a line containing just `+`.


jhernand:
FYI: @cben

PR #115, opened 2020-02-25
Add target to generate _OpenAPI_

jhernand:
This patch adds a new `openapi` target that generates the _OpenAPI_
specification of the model. This is intended to simplify checking
manually that the generated _OpenAPI_ specification is correct.


jhernand:
FYI: @cben

cben:
LGTM.  Works nicely.

PR #128, opened 2020-03-13
Add new fields to add-on and add-on installations

vkareh:
This adds missing fields to the add-on installation struct. It also adds a new add-on field to track the operator name.


vkareh:
@igoihman - please take a look

igoihman:
lGTM

PR #130, opened 2020-03-18
Add a flag to indicate if cluster admin access is enabled

zgalor:


zgalor:
@tzvatot @igoihman PTAL

igoihman:
LGTM

PR #132, opened 2020-03-19
LDAP: Fix attributes property name

vkareh:
When creating an LDAP IDP, the attributes is a required property.
However, the current name resolves in the SDK as "ldap_attributes", thus
failing creation with a "400: 'attributes' field is mandatory" error
message.


vkareh:
@igoihman @zgalor - please take a look

igoihman:
LGTM

PR #135, opened 2020-03-20
Add Cluster API listening method

igoihman:
In order to support Private OSD clusters, we need to mark the API as external or internal.

### DIFF
--- a/model/clusters_mgmt/v1/listening_method_type.model
+++ b/model/clusters_mgmt/v1/listening_method_type.model
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2019 Red Hat, Inc.
+Copyright (c) 2020 Red Hat, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


igoihman:
@vkareh @tzvatot PTAL

model/clusters_mgmt/v1/listening_method_type.model
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+// Cluster components listening method.
+enum ListeningMethod {
+	// Uses only internal traffic.
+	ListeningMethodInternal
+
+	// Uses both external and internal traffic.
+	ListeningMethodExternal
```
vkareh: These should be `Internal` and `External`, or else when using the SDK you end up with the request being sent as
```json
{
  ...
  "listening": "listening_method_external"
}
```

zvikorn:
LGTM

zgalor:
What is it used for?

igoihman:
@zgalor for private cluster support

tzvatot:
Indentation seems wrong. Other than that LGTM.

igoihman:
@tzvatot I fixed it, merging the change.

PR #139, opened 2020-03-24
[SDB-938] Update AMS Models

gshilin-sdb:
https://issues.redhat.com/browse/SDB-938

PR #140, opened 2020-03-24
add sockets to cluster_metrics_type

petli-openshift:


petli-openshift:
These attributes are missing, and it blocks for my currently task (relatively urgent).
```
$ ocm get cluster 1c3vfrvnsg7fcatdbu2k5egqbeuogp3s
"metrics": {
    ......
    "sockets": {
      "updated_timestamp": "2020-03-24T17:39:59.192182Z",
      "used": {
        "value": 0,
        "unit": ""
      },
      "total": {
        "value": 1,
        "unit": ""
      }
    },
    ......
    "operating_system": "Red Hat Enterprise Linux CoreOS",
  },

```

model/clusters_mgmt/v1/cluster_metrics_type.model
```diff
@@ -35,4 +38,7 @@ struct ClusterMetrics {
 
 	// The number of nodes provisioned for the cluster.
 	Nodes ClusterNodes
+
+    // The operation system provisioned for the cluster.
+	OperatingSystem String
```
vkareh: Do we want to make this a separate enum type? We only support `Red Hat Enterprise Linux` and `Red Hat Enterprise Linux CoreOS`, whereas a string would allow arbitrary values. Not a blocker.

vkareh:
@igoihman - ping

model/clusters_mgmt/v1/cluster_metrics_type.model
igoihman: We already have a cluster operating system type, is it something you can use here?

petli-openshift: Yep, good point. I can update it. just to confirm, I happened to review this from @gshilin-sdb ,
https://github.com/openshift-online/ocm-api-model/pull/139/files#diff-bcc75eb265fd6b0141bc43bafaeaedf6R17

dose the enum break right now?

petli-openshift: I remove `OperatingSystem`. I am sure I m not using it now.

PR #145, opened 2020-03-30
Add ClusterAlertsFiring type and field in ClusterMetrics

nimrodshn:
cc: @igoihman @tzvatot @cben PTAL

### DIFF
--- a/model/clusters_mgmt/v1/cluster_alerts_firing_type.model
+++ b/model/clusters_mgmt/v1/cluster_alerts_firing_type.model
@@ -1,0 +1,30 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/cluster_alerts_firing_type.model
```diff
@@ -0,0 +1,30 @@
...
+
+	// Number of "warning" alerts firing.
+	Warning Integer
+
+	// Number of "high" alerts firing.
+	High Integer
+
+	// Number of "critical" alerts firing.
+	Critical Integer
+}
```
igoihman: no newline

nimrodshn: Thanks, fixed!

PR #147, opened 2020-04-01
Products: Support for top-level cluster types

vkareh:
We want to support the concept of product types for clusters, rather
than relying on discrete flags (e.g. cluster.Managed). This will allow
us to enforce specific configurations depending on the product (e.g.
automatically adding necessary add-ons to an RHMI cluster), as well as
improved quota reservation mechanisms (e.g. bypassing quota checks when
reshaping MOA clusters).


igoihman:
LGTM

PR #148, opened 2020-04-02
Products: Add product attribute to cluster object

vkareh:
To support creating clusters of a specific product type, we need to
allow the cluster object to reference the product object itself. When
a cluster is created as part of a specific product, certain defaults
will be enforced and quota will be reserved accordingly.


vkareh:
This builds upon https://github.com/openshift-online/ocm-api-model/pull/147 to fully support product types for clusters.

PR #153, opened 2020-04-06
Fix pull secret deletion path

rawsyntax:
This is a fix for the code I submitted in https://github.com/openshift-online/ocm-api-model/pull/149


rawsyntax:
@vkareh PTAL

vkareh:
Agh, my bad that I didn't catch the plural resources in #149

This looks good now :crossed_fingers:

PR #156, opened 2020-04-07
Add HealthState field to Cluster type.

nimrodshn:
cc: @igoihman @vkareh

### DIFF
--- a/model/clusters_mgmt/v1/cluster_health_state_type.model
+++ b/model/clusters_mgmt/v1/cluster_health_state_type.model
@@ -1,0 +1,27 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// ClusterHealthState indicates the health of a cluster.
+enum ClusterHealthState {
+  // Cluster is Ready and healthy.
+  Healthy   
+
+  // Cluster is Ready and unhealthy.
+  Unhealthy 
+
+  // Cluster health is unkown.
+  Unknown
+}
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/cluster_health_state_type.model
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// ClusterHealthState indicates the health of a cluster.
+enum ClusterHealthState {
+    // Cluster is Ready and healthy.
+    ClusterHealthStateHealthy
```
vkareh, line 20: Indentation looks off here /nitpick

model/clusters_mgmt/v1/cluster_health_state_type.model
```diff
@@ -0,0 +1,27 @@
...
+enum ClusterHealthState {
+    // Cluster is Ready and healthy.
+    ClusterHealthStateHealthy   
+
+    // Cluster is Ready and unhealthy.
+	ClusterHealthStateUnhealthy 
+
+    // Cluster health is unkown.
+	ClusterHealthStateUnknown
+}
```
vkareh, line 27: I think these health states should just be the actual value, like
```
enum ClusterHealthState {
  Healthy
  Unhealth
  Unknown
}
```
The reason for that is that the metadata generator for the SDK will take the entry from this file and auto-generate the allowed values for state, which in this case will end up looking like this:
```
{
  "health_state": "cluster_health_state_healthy"
}
```

nimrodshn: Ah ok cool

nimrodshn: Done.

model/clusters_mgmt/v1/cluster_health_state_type.model
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// ClusterHealthState indicates the health of a cluster.
+enum ClusterHealthState {
+  // Cluster is Ready and healthy.
+  Healthy
```
vkareh, line 20: indentation still an issue :smile: 
I think you have multiple spaces here, where the others have a single tab

nimrodshn: maybe now?? :man_facepalming:

PR #157, opened 2020-04-07
Update to metamodel 0.0.27

vkareh:


vkareh:
@igoihman - just for consistency, this shouldn't change anything

PR #163, opened 2020-04-19
Adding register_disconnected to api model

zvikorn:

### DIFF
--- a/model/clusters_mgmt/v1/register_disconnected_resource.model
+++ b/model/clusters_mgmt/v1/register_disconnected_resource.model
@@ -1,0 +1,31 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Define the registered disconnected cluster.
+resource RegisterDisconnected {
+
+	// Adds (registers) a disconnected cluster.
+	method Add {
+		// Details of the disconnected cluster.
+		in out Body RegisterDisconnected
+	}
+
+	// Reference to the resource that manages a specific disconnected cluster.
+	locator RegisterDisconnected {
+		target RegisterDisconnected
+		variable ID
+	}
+}
--- a/model/clusters_mgmt/v1/register_disconnected_type.model
+++ b/model/clusters_mgmt/v1/register_disconnected_type.model
@@ -1,0 +1,28 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


zvikorn:
@vkareh Please let me know if this is the correct way to do that. Also, How do I know that I am not missing anything in the register_disconnected_type.model file (any other attribute)? I guess that the fields I can add to the body, can be taken or seen from the telemetry? need some clarification here please...

vkareh:
This is a tricky one, as it breaks with the "resource" paradigm in our API. Doing `POST /api/clusters_mgmt/v1/register_disconnected` implies that you are creating a resource of type "RegisterDisconnected", which is not the case. You are just creating a Cluster resource, it just happens to follow a different internal flow.

Not that I want to break the API, but I wonder if we should refactor our `/clusters` endpoints and allow for something like
```
POST /.../clusters --body {..., "product": {"id": "ocp"}, "disconnected": true}
```
This would mean that the cluster handler will redirect the flow to whatever the service.RegisterDisconnected does when it finds a "disconnected OCP" payload. Then we can keep `/register_disconnected` as a legacy alias of the new flow and not have to include it in this repository.

zvikorn:
@igoihman @tzvatot @zgalor @nimrodshn need your thoughts on Victor's comment.

tzvatot:
I think "register_disconnected" was not suppose to be added this way. It should have followed the paradigm that @vkareh pointed out.

I think it wrongly followed the logic of "register_cluster" that is called by AMS, have a different payload and credentials.

If we like to fix the "register_disconnected" endpoint to align with our conventions, then I agree we should fix it first and then expose it properly in the API model.

Is there a strong case to add it to the model as it is?

@zgalor WDYT?

zgalor:
We had this discussion when the feature was born.
The problem was the the "disconnected" field is store in the AMS and not CS.
I discussed the following options with @tzvatot and @jhernand:
1. Expose it as an API field but not as a model field. This means the user will not see it when they GET the cluster, because it is kept in AMS. not very RESTful.
2. Add this field as a parameter to POST. (No one liked it).
3. Add an action like `register` that allows POST only and receives a `Cluster` object but handles it differently than `/clusters` endpoint.

@jhernand didn't really like the third option as it breaks the Collections model (I think this was before the ocm_api_model project, but still), but since we had the precedent of the `register` endpoint,  he agreed it was the lesser of evils.

I guess the ocm_api_model doesn't expose the `register` endpoint because it's internal, so we either need to refactor the model to support actions like `register_disconnected`, or change back to options 1 or 2...

tzvatot:
@zgalor if this is an internal API, why are we exposing it in the API model?

zgalor:
I didn't know we expose `register`. But if we are, we can expose `register_disconnected` and follow the same pattern

tzvatot:
After talking with @zgalor :

1. The "register_disconnected" API is exposed to end users. Hence we must add it also in the model here.
2. The fact that it's implemented the way it is in the backend, is irrelevant to the discussion here. We should expose it as it is implemented now. If we change the backend implementation later on, we should go and fix the API model accordingly.
3. The reasons that the backend implementation is as it is today haven't changed, so I don't see any reason to revisit them now. We can revisit if anyone thinks otherwise.

@zgalor @vkareh @zvikorn make sense?

cben:
> if this is an internal API, why are we exposing it in the API model?

Besides this particular API being relevant to end users, this model is used to generate client and server code, which can be useful even for "internal" account-manager<->clusters-service APIs.
Also, our team + QE + others in Red Hat do benefit from having OpenAPI documentation for all APIs, including "internal".
So I think our goal here is cover all APIs.

When APIs are actually internal (by requiring priviledged auth that external users will never have), best we can do with current architecture is note that in their description.

zvikorn:
@vkareh @tzvatot @zgalor now when we decided to proceed as is, Could you PTAL?

zgalor:
Again, as @vkareh explained - this will not work. We need to refactor the model (and probably code generation part) to support actions. Meaning sending `Cluster` type object to `register_disconnected` endpoint

PR #165, opened 2020-04-22
Added HostPrefix To Network

pvasant:


model/clusters_mgmt/v1/network_type.model
```diff
@@ -24,4 +24,6 @@ struct Network {
 
 	// IP address block from which to assign service IP addresses, for example `172.30.0.0/16`.
 	ServiceCIDR String
+
+	HostPrefix Integer
```
vkareh, line 28: Please add a comment above this line, following the format of the other attributes. These are parsed and added to the openapi docs

vkareh: Please follow the format above for indentation and spacing

vkareh: Okay, @zvikorn had the same issue - your IDE (probably goland, right?) doesn't believe you that this is a Go file (because it's not, really), so it doesn't want to apply similar formatting as the rest of the file for some reason.

In your patch, the comment is prefixed by 4 spaces, but it should be 1 tab instead. (the HostPrefix line seems fine for now)

pvasant: @vkareh  Yes I use goland and i was using tab not spaces. I edited it directly in this GitHub editor to resolve.

vkareh:
Closed in favor of https://github.com/openshift-online/ocm-api-model/pull/167

PR #169, opened 2020-04-23
Add organization create

rawsyntax:
I generated the sdk locally and tested this endpoint

### DIFF
--- a/model/accounts_mgmt/v1/organization_resource.model
+++ b/model/accounts_mgmt/v1/organization_resource.model
@@ -26,6 +26,11 @@ resource Organization {
 		in out Body Organization
 	}
 
+	// Creates the organization
+	method Add {
+		in out Body Organization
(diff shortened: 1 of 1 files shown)


rawsyntax:
@igoihman @vkareh PTAL

vkareh:
@rawsyntax - what's your use-case here? The `organizations_resource.model` already has this same method, so merging this would mean that you can add... an organization inside another organization?

https://github.com/openshift-online/ocm-api-model/blob/master/model/accounts_mgmt/v1/organizations_resource.model#L52-L55

current:
* `POST /api/accounts_mgmt/v1/organizations` 

this patch:
* `POST /api/accounts_mgmt/v1/organizations/123456`

rawsyntax:
My mistake here. I missed that organization create was already in the sdk

PR #184, opened 2020-05-25
Add parameters 'offset' and 'tail' to log resource

zgalor:


zgalor:
@igoihman @tzvatot PTAL

PR #189, opened 2020-06-04
Support for all kinds of Labels

gshilin-sdb:


model/accounts_mgmt/v1/account_label_resource.model
```diff
@@ -0,0 +1,38 @@
...
+
+// Manages a specific account label.
+resource AccountLabel {
+	// Retrieves the details of the label.
+	method Get {
+		out Body AccountLabel
+	}
+
+	// Create a new account label.
+	method Add {
```
vkareh, line 25: The `Add` method should be part of the `account_labels_resource`.

model/accounts_mgmt/v1/account_type.model
```diff
@@ -26,5 +26,5 @@ class Account {
 	UpdatedAt Date
 	ServiceAccount Boolean
 	Organization Organization
-	Labels []Label
+	Labels []AccountLabel
```
vkareh, line 29: Does the `Account` model have the `Labels` directly or as a link? In the `clusters_mgmt` API, when a resource contains a list of another resource, we typically set it as a link. For example:
```
{
  "kind": "Account",
  "href": "/api/accounts_mgmt/v1/account/###",
  "labels": {
    "kind": "AccountLabelsLink",
    "href": "/api/accounts_mgmt/v1/account/###/account_labels"
  }
}
```
In most cases we don't expose the `"items"` directly on the parent resource and let the client request those manually as needed.

If this is (or should be) the case, then this line should change to
```
link Labels []AccountLabel
```

model/accounts_mgmt/v1/label_type.model
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2019 Red Hat, Inc.
+Copyright (c) 2020 Red Hat, Inc.
```
vkareh, line 2: No need to update copyright year on existing files (unless the 2019 was a mistake and the file was actually created in 2020).

model/accounts_mgmt/v1/organization_label_resource.model
```diff
@@ -0,0 +1,38 @@
...
+
+// Manages a specific organization label.
+resource OrganizationLabel {
+	// Retrieves the details of the label.
+	method Get {
+		out Body OrganizationLabel
+	}
+
+	// Create a new organization label.
+	method Add {
```
vkareh, line 25: The `Add` method should be part of `organization_labels_resource`

model/accounts_mgmt/v1/labels_resource.model
```diff
@@ -0,0 +1,50 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the collection of labels.
+resource Labels {
```
vkareh, line 18: Is this a generic read-only list of _all_ labels (regardless or whether they belong to an Account, Organization, or Subscription)? Wondering since there are no other methods defined.

model/accounts_mgmt/v1/subscription_label_resource.model
```diff
@@ -0,0 +1,38 @@
...
+
+// Manages a specific subscription label.
+resource SubscriptionLabel {
+	// Retrieves the details of the label.
+	method Get {
+		out Body SubscriptionLabel
+	}
+
+	// Create a new subscription label.
+	method Add {
```
vkareh, line 25: The `Add` method should be part of `subscription_labels_resource`.

model/accounts_mgmt/v1/account_type.model
gshilin-sdb: Labels are direct part of Account/Organization/Subscription

model/accounts_mgmt/v1/label_type.model
gshilin-sdb: This file was created couple of weeks ago, so 2020

model/accounts_mgmt/v1/account_type.model
vkareh: Got it, thanks

model/accounts_mgmt/v1/label_type.model
vkareh: Thanks for fixing it then :)

model/accounts_mgmt/v1/labels_resource.model
gshilin-sdb: Exactly, all labels

vkareh: I like it

gshilin-sdb:
Will also try to remove some duplication, but cannot promise...

gshilin-sdb:
Removed duplication

vkareh:
Very nice! Let me see if I understand this - the `Label` resource refers to a read-only label at the root level, so I can query for all labels. Then the `GenericLabel` is an abstraction that allows Accounts, Orgs, Subs to have Labels attached to them (with all the CRUD operations). Does that sound right?

gshilin-sdb:
Exactly. And the Label resource also returns GenericLabels

vkareh:
/lgtm

I ran against the ocm-sdk-go build scripts and this is a portion of the OpenAPI I get:
```diff
+++ b/openapi/accounts_mgmt/v1/openapi.json
@@ -115,6 +115,14 @@
       "get": {
         "description": "Retrieves the list of accounts.\n",
         "parameters": [
+          {
+            "name": "fetchlabels_labels",
+            "description": "If true, includes the labels on an account in the output. Could slow request response time.\n",
+            "in": "query",
+            "schema": {
+              "type": "boolean"
+            }
+          },
```
That `fetchlabels_labels` is a bit concerning, as it should be `fetchLabels`, but this is probably the ocm-api-metamodel's fault. I'm not blocking this patch over that, but let's keep an eye on that in case that spec causes issues, then we can fix it in the metamodel.

vkareh:
Thanks again, and apologies for the delayed code review.

PR #194, opened 2020-06-17
cluster: Remove support for expiration_timestamp

vkareh:
This option still exists but will no longer be supported. Clusters in
production never expire automatically.


oourfali:
lgtm

PR #200, opened 2020-06-25
Add syncsets API

zgalor:


model/clusters_mgmt/v1/syncset_type.model
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of a syncset.
+class Syncset {
+	// List of k8s objects to configure for the cluster.
+	Resources []Interface
```
vkareh, line 20: We should probably also return the ID

vkareh: ah, ID is set by default by the generator - duh :stuck_out_tongue:

vkareh:
Some relevant SDK output when building with openshift-online/ocm-api-metamodel/pull/90

`clustersmgmt/v1/syncset_type.go`:
```go
// Representation of a syncset.
type Syncset struct {
	id        *string
	href      *string
	link      bool
	resources []interface{}
}

// ...snip...

// Resources returns the value of the 'resources' attribute, or
// the zero value of the type if the attribute doesn't have a value.
//
// List of k8s objects to configure for the cluster.
func (o *Syncset) Resources() []interface{} {
	if o == nil {
		return nil
	}
	return o.resources
}

// GetResources returns the value of the 'resources' attribute and
// a flag indicating if the attribute has a value.
//
// List of k8s objects to configure for the cluster.
func (o *Syncset) GetResources() (value []interface{}, ok bool) {
	ok = o != nil && o.resources != nil
	if ok {
		value = o.resources
	}
	return
}
```
`openapi/clusters_mgmt/v1/openapi.json`:
```json
      "Syncset": {
        "description": "Representation of a syncset.\n",
        "properties": {
          "kind": {
            "description": "Indicates the type of this object. Will be 'Syncset' if this is a complete object or 'SyncsetLink' if it is just a link.\n",
            "type": "string"
          },
          "id": {
            "description": "Unique identifier of the object.\n",
            "type": "string"
          },
          "href": {
            "description": "Self link.\n",
            "type": "string"
          },
          "resources": {
            "description": "List of k8s objects to configure for the cluster.\n",
            "type": "array",
            "items": {
              "type": "object"
            }
          }
        }
      },
```

This is looking good! :+1: 
We might want to test this a bit before merging openshift-online/ocm-api-metamodel/pull/90

zgalor:
@vkareh @igoihman what is required for this to pass the build?
A new SDK?

vkareh:
Correct. We need a new SDK release with the changes from the metamodel. Otherwise the test is being run without the SDK knowing about the Interface type

igoihman:
@zgalor I updated the metamodel version in `api-model` and in the sdk. please rebase.

zgalor:
@igoihman done.
@vkareh @igoihman PTAL

PR #203, opened 2020-06-29
Change feature toggle to POST with request context containing org id

tzvatot:
@igoihman @tiwillia @markturansky @rawsyntax @zgalor PTAL

### DIFF
--- a/model/accounts_mgmt/v1/feature_toggle_resource.model
+++ b/model/accounts_mgmt/v1/feature_toggle_resource.model
@@ -16,8 +16,9 @@ limitations under the License.
 
 // Manages feature toggles.
 resource FeatureToggle {
-	// Retrieves the details of the feature toggle.
-	method Get {
-		out Body FeatureToggle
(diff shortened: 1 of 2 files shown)


model/accounts_mgmt/v1/feature_toggle_resource.model
```diff
@@ -16,8 +16,9 @@ limitations under the License.
 
 // Manages feature toggles.
 resource FeatureToggle {
-	// Retrieves the details of the feature toggle.
-	method Get {
-		out Body FeatureToggle
+	// Retrieves the details of the feature toggle by providing query context
+	method Post {
+		in Request FeatureToggleQueryRequest
```
zgalor, line 21: This is not very restful. Why not add the Organization Id to the `FeatureToggle` type?

tzvatot: Because the organization id is not related to the response.

The request need to contain some context. In this case - the org id. The response however, is true/false.

PR #207, opened 2020-07-01
Change the feature toggle API to be /feature_toggles/id/query using POST with org id as context

tzvatot:
@nimrodshn @zgalor @vkareh PTAL

### DIFF
--- a/model/accounts_mgmt/v1/feature_toggle_query_resource.model
+++ b/model/accounts_mgmt/v1/feature_toggle_query_resource.model
@@ -14,9 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Manages feature toggle.
-resource FeatureToggle {
-	locator Query {
-		target FeatureToggleQuery
+// Manages feature toggle query.
+resource FeatureToggleQuery {
+	// Retrieves the details of the feature toggle by providing query context
+	method Post {
+		in Request FeatureToggleQueryRequest
+		out Response FeatureToggle
 	}
 }
--- a/model/accounts_mgmt/v1/feature_toggle_query_request_type.model
+++ b/model/accounts_mgmt/v1/feature_toggle_query_request_type.model
@@ -14,9 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Manages feature toggle.
-resource FeatureToggle {
-	locator Query {
-		target FeatureToggleQuery
-	}
+struct FeatureToggleQueryRequest {
+	OrganizationID String
 }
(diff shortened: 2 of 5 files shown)


model/accounts_mgmt/v1/feature_toggle_query_resource.model
```diff
@@ -0,0 +1,24 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages feature toggle query.
+resource FeatureToggleQuery {
+	// Retrieves the details of the feature toggle by providing query context
```
vkareh, line 19: This is a `POST` request, but "Retrieves"? The generated SDK will lists it as responding with "201 - Created"

model/accounts_mgmt/v1/feature_toggle_query_resource.model
```diff
@@ -0,0 +1,24 @@
...
+	method Post {
+		in Request FeatureToggleQueryRequest
```
vkareh, line 21: Ah okay, I see you're "posting" a query request and getting back a toggle setting? If `FeatureToggleQueryRequest` only has an Org ID, why can't we `GET /api/.../{org_id}/feature_toggle/{feature_toggle_id}`, like we do for all other resources?

model/accounts_mgmt/v1/feature_toggle_query_resource.model
tzvatot: Oh great, I was wondering about it. How can I change it to be 200?

model/accounts_mgmt/v1/feature_toggle_query_resource.model
tzvatot: Feature toggle can have many parameters as input.
For example, we can decide to toggle feature by cluster, by region, or anything else.
Right now our only use case is for org.

model/accounts_mgmt/v1/feature_toggle_query_resource.model
vkareh: It's automatically generated through the metamodel. I don't think we have a way of coercing the response code to be different.

Only `GET` requests respond with 200...

model/accounts_mgmt/v1/feature_toggle_query_resource.model
vkareh: interesting... AMS has an interesting way of dealing with these arbitrary data types. They usually have a `--fetchFooBarData` param in their GET requests. Is this something we can do? For example `GET /../orgs/{org_id} --fetchFeatureToggles`. WDYT?

PR #214, opened 2020-07-10
add register_cluster resource

petli-openshift:


petli-openshift:
@vkareh can we expose this api?

model/clusters_mgmt/v1/register_cluster_request_type.model
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 // Registration of a new cluster to the service.
-struct ClusterRegistration {
+struct RegisterClusterRequest {
```
vkareh, line 18: This is an API-breaking change - it changes existing paths in the OpenAPI spec and renames an entire type in the SDK.

petli-openshift: will it do if I change the name back? my original thought was to make the naming a bit more consistent with other requests. also its endpoint resource (also added in the MR) is missing. doubt it's been used anywhere.

vkareh: You're right about the root-level endpoint, it's new, so won't break anything. Regarding the naming, I fully agree with the rename. So that leaves anything using the SDK (most likely some pipeline somewhere) could break if they upgrade to the new version that contains this change. Assuming this class/type is even used anywhere :stuck_out_tongue: 

@igoihman @tzvatot @zgalor - any opinions about this?

tzvatot: If it's been used, than it's only internally due to the required permissions.
I'm ok with breaking it for consistency. Just FYI for @zgalor @nimrodshn @zvikorn @cben @pvasant: if you bump the sdk version and see failures around integration tests - note you'll need to change the name of the struct.

zgalor: I'm ok with renaming - I don't think it even requires a change in CS code to support this (Tough we should for consistency)

petli-openshift: @vkareh I updated and rebased this MR.

petli-openshift: pls let me know if we can merge this. I really like the way how ocm-sdk builds requests and want to update a client code using that. However, it is not absolutely needed. I can switch to the other alternative if you feel this change is too risky. thanks

PR #215, opened 2020-07-13
Implement Batch Patch Ingresses API endpoint

zvikorn:

https://issues.redhat.com/browse/SDA-2625

### DIFF
--- a/model/clusters_mgmt/v1/ingresses_resource.model
+++ b/model/clusters_mgmt/v1/ingresses_resource.model
@@ -37,6 +37,11 @@ resource Ingresses {
 		in out Body Ingress
 	}
 
+	// Updates all ingresses
+	method Update {
+		in out Body []Ingress
+    }
+
 	// Reference to the service that manages a specific ingress.
 	locator Ingress {
 		target Ingress


vkareh:
uh, just read the Jira thread and I want to make sure we understand the implications of doing this:
* `PATCH` is atomic, so either _all_ the ingresses in the request get patched, or none do. Let's make sure the server does this
* How does patch "unset" an ingress? What's the plan for that? Since you need an ID to do that, what does the request look like? In a normal patch request, you can do: `{ "id": "foo", "name": null }` to unset the `name` attribute, how can we do that for an entire ingress object?

zvikorn:
@vkareh I guess we cannot unset an ingress with PATCH. We still have to call DELETE. We will iterate all ingress exist in the body, and update one by one. Does this make sense?

vkareh:
Well, what is the proposed payload? Not sure what you have in mind, but I was envisioning something like:
```
PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses << EOF
{
  "ingresses": [
    {"id": "apps", "ingress": {
      "listening": "internal"
    }},
    {"id": "apps2", "ingress": null}
  ]
}
EOF
```
In this case, the `apps` ingress gets set to private, and the `apps2` ingress is removed as it contains the entry, but without a body.

But it all depends on what you had in mind for the implementation...

zvikorn:
@vkareh  I was thinking about:
PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses << EOF
{
  "ingresses": [
             { "id": "apps", 
               "listening": "internal"
             },
             {"id": "apps2", 
              "listening": "external"
             }
  ]
}
EOF

Without the deletion option

vkareh:
I like how clean that is, but in `PATCH` you typically only send the attributes that you're updating in the payload, so separating the `id` from the rest of the body is probably a good idea (and thus allows for deletion). Although patches are idempotent, so it shouldn't matter much.

I guess we're treating this as a "batch" operation, rather than a straight-up "patch", so in this case what we want is to consolidate the two following requests:
* `PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses/apps << {"listening": "internal"}`
* `PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses/apps2 << {"listening": "external"}`
so that they are a single call:
```
PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses << EOF
{
  "apps": {"listening": "internal"},
  "apps2": {"listening": "external"}
}
EOF
```
or
```
PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses << EOF
[
  {"id": "apps", "ingress": {"listening": "internal"}},
  {"id": "apps2", "ingress": {"listening": "external"}}
]
EOF
```
Now I'm thinking if sending an array, rather than an object, is possible...

Also leads me to think whether we want a batch API:
```
POST /api/clusters_mgmt/v1/clusters/{cluster_id}/batch << EOF
{
  "items": [
    {"op": "patch", "resource": "ingress", "id": "apps", "body": {"listening": "internal"}},
    {"op": "patch", "resource": "ingress", "id": "apps2", "body": {"listening": "external"}}
  ]
}
EOF
```
which would allow us to do all sorts of crazy things with a cluster, like setting groups and users at the same time or installing add-ons while updating CIDRs and stuff like that... and that's where my bikeshedding ends. :no_bicycles: :laughing:

zvikorn:
:-) nice :-) I think the batch API is a great idea but it's something we will have to have a jira card for. However, I am not sure we need something like this as I do not think we have use cases for bulk operations (but you know better than me). But as I said we can start a discussion on this, involving the right engineers. Anyway, what do you think then about doing the following for now (this feature)?
PATCH /api/clusters_mgmt/v1/clusters/{cluster_id}/ingresses << EOF
[
  {"id": "apps", "ingress": {"listening": "internal"}},
  {"id": "apps2", "ingress": {"listening": "external"}}
]
EOF

for this feature?

vkareh:
> something we will have to have a jira card for

Already in my pipeline since April ;)

Regarding this PR, I'm fine with either that or your original one, honestly. However, I'm getting build errors when generating the SDK:
```
E: Type of parameter 'ClustersMgmt.V1.Ingresses.Update.Body' should be scalar or struct but it is 'list'
E: Method 'ClustersMgmt.V1.Ingresses.Update' should have exactly one struct parameter but it has 0
E: Type of parameter 'ClustersMgmt.V1.Ingresses.Update.Body' should be scalar or struct but it is 'list'
E: Method 'ClustersMgmt.V1.Ingresses.Update' should have exactly one struct parameter but it has 0
E: Can't read model: there were 4 errors
make: *** [Makefile:54: generate] Error 1
```

zvikorn:
I assume there is somewhere a defintion for the update operation that allow either struct or scalar only, and we need to add the option of List

zvikorn:
So this error is from metamodel/pkg/language/checks.go -> unc (r *Reader) checkUpdate(method *concepts.Method) 
Do we want to change this to support list?

vkareh:
>  Do we want to change this to support list?

I would assume yes, is this something you can start work on?

zvikorn:
sure

zvikorn:
@vkareh Will this hopefully be built once we merge https://github.com/openshift-online/ocm-api-metamodel/pull/92?

vkareh:
Created https://github.com/openshift-online/ocm-api-model/pull/227 which is this same change but using the ocm-api-metamodel v0.0.31, which supports patching list endpoints

PR #219, opened 2020-07-20
Introduce API for getting cluster's provision shard

tzvatot:
@zgalor @vkareh @igoihman @nimrodshn PTAL


model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -60,6 +60,10 @@ resource Cluster {
 		target ExternalConfiguration
 	}
 
+	// Reference to the resource that manages the external configuration.
```
zgalor, line 63: `external cluster services`

model/clusters_mgmt/v1/external_cluster_service_resource.model
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages ExternalClusterService.
+resource ExternalClusterServices {
```
zgalor, line 18: ExternalClusterService - single object. Or rename the file.
I think the metamodel will have a problem with giving a resource that is not a collection a plural name

model/clusters_mgmt/v1/external_cluster_services_resource.model
```diff
@@ -0,0 +1,38 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the collection of external cluster services.
+resource ExternalClusterServicesCollection {
```
zgalor, line 18: I see what you did here, but I'm not sure openapi.yaml generation would be as you expect

model/clusters_mgmt/v1/root_resource.model
```diff
@@ -61,4 +61,9 @@ resource Root {
 	locator Products {
 		target Products
 	}
+
+	// Reference to the resource that manages the collection of external cluster services.
+	locator ExternalClusterServices {
```
zgalor, line 66: this can be accessed both from root and from the cluster?

model/clusters_mgmt/v1/external_cluster_service_resource.model
tzvatot: It's working. I tested it.
Need to fix the comment though.

Per the filename, there's a problem because the name "external_cluster_services" is the single object here. 
And there's a list of "external_cluster_services", which I called "ExternalClusterServicesCollection".
So I will rename the later filename.

model/clusters_mgmt/v1/external_cluster_services_resource.model
tzvatot: it looks as I expected

model/clusters_mgmt/v1/external_cluster_service_resource.model
tzvatot: Fixed. PreTAL

model/clusters_mgmt/v1/root_resource.model
tzvatot: yes. The root is a general config.
And the cluster is link to the one attached to the cluster.
Kind of like flavour. Only flavour is exposed in the cluster body, and this isn't.

model/clusters_mgmt/v1/external_cluster_service_resource.model
zgalor: so the endpoint will be `/external_cluster_services` both for root and for cluster?
It is confusing and nonunified. All our other plural endpoints indicate collection.
Maybe we should consider renaming it to something like `external_server_config` (that can be plural `configs` if calling from root)

tzvatot: It is confusing, I agree.
It was a bad naming decision on my behalf. I shouldn't have called that single tuple of [hive+awsAcctOperator+gcpProjOperator] something that is plural.

That said, it's already out there. If I want to rename it, I need to do a huge change, which includes configuration changes both on CS basecode, and app-interface. 

I can mitigate that discrepancy by calling the collection of "external-cluster-services" as "external_cluster_services_list", but this also isn't consistent with the other endpoints that we have.

Another option would be to drop the root resource all together, and not expose it at all. You can still see the configuration by querying the DB. And SRE will still be able to see which one is used for their cluster, but won't be able to see all of them using the API.

WDYT?

zgalor: I don't think it's such a big change. You don't need to change everything. Since it's the first time it's exposed in the API - you can determine the way it's called. Maybe `external_cluster_services_config` is not so far from current naming

model/clusters_mgmt/v1/server_config_type.model
```diff
@@ -0,0 +1,21 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of a server config
+class ServerConfig {
```
zgalor, line 18: Is this a place holder for other potential fields?

tzvatot: Yes, SRE may want to see more info than just the server URL.

PR #221, opened 2020-07-21
Expose if a region supports multi AZ

tzvatot:
@nimrodshn @zgalor @zvikorn  PTAL

### DIFF
--- a/model/clusters_mgmt/v1/cloud_region_type.model
+++ b/model/clusters_mgmt/v1/cloud_region_type.model
@@ -30,4 +30,7 @@ class CloudRegion {
 
 	// Whether the region is enabled for deploying an OSD cluster.
 	Enabled Boolean
+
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/cloud_region_type.model
```diff
@@ -30,4 +30,7 @@ class CloudRegion {
 
 	// Whether the region is enabled for deploying an OSD cluster.
 	Enabled Boolean
+
+	// Whether the region supports multi availability zones.
```
zgalor, line 34: np: multiple

zgalor:
LGTM. see np

model/clusters_mgmt/v1/cloud_region_type.model
tzvatot: fixed

PR #224, opened 2020-07-24
removing 'deprovision' suffix from logs endpoint

zvikorn:

https://issues.redhat.com/browse/SDA-2686


zvikorn:
@vkareh PTAL

vkareh:
hmm... I think the entire locator object should be removed, otherwise the result is this:
```diff
diff --git a/openapi/clusters_mgmt/v1/openapi.json b/openapi/clusters_mgmt/v1/openapi.json
index 20c92c5..bc11384 100644
--- a/openapi/clusters_mgmt/v1/openapi.json
+++ b/openapi/clusters_mgmt/v1/openapi.json
@@ -2920,7 +2920,7 @@
         }
       }
     },
-    "/api/clusters_mgmt/v1/clusters/{cluster_id}/logs/uninstall/{log_id}": {
+    "/api/clusters_mgmt/v1/clusters/{cluster_id}/logs/uninstall/log": {
       "get": {
         "description": "Retrieves the details of the log.\n",
         "parameters": [
@@ -2931,13 +2931,6 @@
               "type": "string"
             }
           },
-          {
-            "name": "log_id",
-            "in": "path",
-            "schema": {
-              "type": "string"
-            }
-          },
           {
             "name": "offset",
             "description": "Line offset to start logs from. if 0 retreive entire log.\nIf offset > #lines return an empty log.\n",
```

vkareh:
Ah, the resource is still a `List` method - needs to be changed to a `Get`

vkareh:
Or just remove the entire `uninstall_logs` resource, you no longer need a list

zvikorn:
removed the list. I think I still need to the locator:
// Manages a collection of install logs.
resource UninstallLogs {
	
	// Returns a reference to the service that manages an specific log.
	locator Log {
		target Log
	}
}

vkareh:
Ah interesting that there's no UninstallLog resource already... so yes, rename to singular so that it matches all other GET resources

vkareh:
hmmm... this last change doesn't work for me. It removes the `/api/clusters_mgmt/v1/clusters/{cluster_id}/logs/uninstall` endpoint and adds a `/api/clusters_mgmt/v1/clusters/{cluster_id}/logs/uninstall/log` one... I think you might need to change the root_resource reference to it (if there's any)

zvikorn:
What if I add the following to Uninstall_log_resource?
// Retrieves the details of the log.
    	method Get {
    		// Line offset to start logs from. if 0 retreive entire log.
    		// If offset > #lines return an empty log.
    		in Offset Integer = 0
    
    		// Returns the number of tail lines from the end of the log.
    		// If there are no line breaks or the number of lines < tail
    		// return the entire log.
    		// Either 'tail' or 'offset' can be set. Not both. 
    		in Tail Integer = 0
    
    		// Retreived log.
    		out Body Log
    	}

zvikorn:
or maybe add/update:
locator Uninstall {
		target log
		}  
instead:
locator Uninstall {
		target UninstallLog
		
	}  
to logs_resource

vkareh:
This works well for me now!

/lgtm

PR #235, opened 2020-08-06
Added SupportCase resource

Lir10:


vkareh:
This works fine when building the SDK, but I'm wondering why you need separate types for the request and response? They are mostly similar, and any difference can be captured in your API validation or presenter.

Lir10:
@vkareh We are going to use it in OSL and also in api-tests in AMS to verify the request/response. Also, the response isn't exactly the same as the request so it's mostly good for us to know looking at the sdk what the response will look like

vkareh:
got it, makes sense

/lgtm

PR #253, opened 2020-09-13
Add key to label_type

zgalor:
ID can't be used as label key, because it can't support keys with prefix and `/` character.
So ID will be generated by OCM and label key will be passed in by the user.


zgalor:
@igoihman @tzvatot @vkareh PTAL

tzvatot:
LGTM

PR #261, opened 2020-09-24
SDA-2952 - Add "hidden" field to AddOn

mikenairn:
JIRA: https://issues.redhat.com/browse/SDA-2952

### DIFF
--- a/model/clusters_mgmt/v1/add_on_type.model
+++ b/model/clusters_mgmt/v1/add_on_type.model
@@ -34,6 +34,9 @@ class AddOn {
 	// Indicates if this add-on can be added to clusters.
 	Enabled Boolean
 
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/add_on_type.model
```diff
@@ -34,6 +34,9 @@ class AddOn {
 	// Indicates if this add-on can be added to clusters.
 	Enabled Boolean
 
+	// Indicates if this add-on is hidden.
```
mikenairn, line 37: Maybe not the best description. :thinking:

vkareh: meh, looks fine to me

PR #264, opened 2020-10-02
idp: Add HTPasswd provider

vkareh:
This provider allows login into a cluster without having to access the
web console. We only allow a single provider with a single account in
it. Clients using this should be advised to use this only as a temporary
admin account until a different IdP is added, at which point they should
remove it.


vkareh:
Related: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2232

PR #267, opened 2020-10-05
Add upgrade policy state to model API

tzvatot:
@igoihman @zgalor PTAL

### DIFF
--- a/model/clusters_mgmt/v1/upgrade_policy_state_resource.model
+++ b/model/clusters_mgmt/v1/upgrade_policy_state_resource.model
@@ -14,24 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Manages a specific upgrade policy.
-resource UpgradePolicy {
-	// Retrieves the details of the upgrade policy.
+// Manages a specific upgrade policy state.
+resource UpgradePolicyState {
+	// Retrieves the details of the upgrade policy state.
 	method Get {
-		out Body UpgradePolicy
+		out Body UpgradePolicyState
 	}
 
-	// Update the upgrade policy.
+	// Update the upgrade policy state.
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/upgrade_policy_state_resource.model
```diff
@@ -0,0 +1,28 @@
...
+
+// Manages a specific upgrade policy state.
+resource UpgradePolicyState {
+	// Retrieves the details of the upgrade policy state.
+	method Get {
+		out Body UpgradePolicyState
+	}
+
+	// Update the upgrade policy state.
+	method Update {
```
zgalor, line 25: How will this be handled in terms of permissions? we don't want the user to be allowed to update the state, only the operator, right?

model/clusters_mgmt/v1/upgrade_policy_resource.model
```diff
@@ -29,4 +29,9 @@ resource UpgradePolicy {
 	// Deletes the upgrade policy.
 	method Delete {
 	}
+
+	// Reference to the state of the upgrade policy.
+	locator State {
```
zgalor, line 34: Don't you want to add it to `update_policy_type.model` as well?

tzvatot: No.

There's a design discussion around it.

nimrodshn:
LGTM :+1:

model/clusters_mgmt/v1/upgrade_policy_state_resource.model
tzvatot: Only the operator should access this endpoint, but due to the auth method we cannot do otherwise.

PR #279, opened 2020-10-12
Remove redundant fields

pvasant:


pvasant:
@vkareh  PTAL I updated UI. It is safe to remove from the API.

PR #281, opened 2020-10-14
flavours: Remove infra and compute nodes

vkareh:
Since infra and compute nodes are now calculated based on a combination
of CCS and MultiAZ flags, they are no longer considered part of
a flavour, since they can be different defaults in each combination of
the flags. We now calculate the required default nodes as part of the
cluster creation process.


vkareh:
This is a follow-up to https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2246

PR #283, opened 2020-10-20
Add RawID field to version_type

zgalor:


zgalor:
@tzvatot @igoihman PTAL

tzvatot:
LGTM

PR #296, opened 2020-11-10
Add Taints field to MachinePool type

nimrodshn:
cc: @igoihman

### DIFF
--- a/model/clusters_mgmt/v1/taint_type.model
+++ b/model/clusters_mgmt/v1/taint_type.model
@@ -1,0 +1,28 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


zvikorn:
LGTM.

model/clusters_mgmt/v1/taint_type.model
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+
+// Representation of a Taint set on a MachinePool in a cluster.
+class Taint {
```
jhernand, line 19: Is this an entity with its own identity? I'd say it isn't, therefore it should be a `struct` instead of a `class`, so that it doesn't automatically have the `kind`, `id` and `href` attributes.

nimrodshn: Correct it isn't. 
Wasn't sure about the details of `struct` vs. `class` semantics..

nimrodshn: @jhernand ptal #297

PR #302, opened 2020-11-24
Add etcd_encryption to sdk

zvikorn:

https://issues.redhat.com/browse/SDA-3362

### DIFF
--- a/model/clusters_mgmt/v1/cluster_type.model
+++ b/model/clusters_mgmt/v1/cluster_type.model
@@ -186,4 +186,8 @@ class Cluster {
 
 	// Node drain grace period.
 	NodeDrainGracePeriod Value
+
+	// Indicates whether that etcd is encrypted or not.
+	// This is set only during cluster creation.
+	EtcdEncryption Boolean
 }


model/clusters_mgmt/v1/cluster_type.model
```diff
@@ -186,4 +186,7 @@ class Cluster {
 
 	// Node drain grace period.
 	NodeDrainGracePeriod Value
+
+	// Indicates whether that etcd is encrypted or not
```
jhernand, line 190: Try to end sentences in comments with punctuation: `Indicates ... or not.`. It isn't very relevant here, or now. But in the future I would like to automatically generate AsciiDoc documentation from these comments, and the tool to do that will most likely use the dot to detect the end of the sentence and to use it as the summary of the documentation for that item.

I'd also suggest to elaborate a bit in the documentation to explain if this is something that can be set only when the cluster is initially created or if it can be updated once the cluster is installed.

model/clusters_mgmt/v1/cluster_type.model
```diff
@@ -186,4 +186,7 @@ class Cluster {
...
+	EtcdEncryption Boolean
```
jhernand, line 191: Is there any chance that we will want or need to add other etcd related attributes in the future? If so I`d suggest to use a struct here:

```
struct Etcd {
  EncryptionEnabled Boolean
}
```

Or even two, one for general etcd things and another one for etcd encryption in particular:

```
struct Etcd {
  Encryption EtcdEncryption
}

struct EtcdEncription {
  Enabled Boolean
}
```

jhernand (commented):
Looks good, with some minor comments.

model/clusters_mgmt/v1/cluster_type.model
zvikorn: I do not think so. @igoihman ?

model/clusters_mgmt/v1/cluster_type.model
zvikorn: Adding the dot. For now, we do it only for creation but we might do it for patching in the future. The question is if it worth adding a comment now for cluster creation and when patching is done, to update this comment..... WDYT?

jhernand: I think it is worth adding and later updating.

model/clusters_mgmt/v1/cluster_type.model
igoihman: the current plan is to have it enabled or disabled. I'm not aware of other attributes we should support

jhernand (approved):
Looks good, thanks @zvikorn.

PR #308, opened 2021-01-15
cluster: Remove ClusterAdminEnabled

vkareh:
Since this opt-in flag has been removed from the API, we remove it also
from the model so that they stay consistent.


vkareh:
@zvikorn please take a look

PR #309, opened 2021-01-18
Add addon requirement type

mikenairn:


model/clusters_mgmt/v1/add_on_requirement_type.model
```diff
@@ -0,0 +1,32 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of an add-on requirement.
+class AddOnRequirement {
```
jhernand, line 18: Is this really a `class` or just a `struct`? Using a class will automatically add the `id` and `href` attributes, assuming that there is an endpoint in the API where these objects can be retrieved. For example:

```
GET /api/clusters_mgmt/v1/add_on_requirements/123
```

But I don't see such a thing.

model/clusters_mgmt/v1/add_on_requirement_type.model
```diff
@@ -0,0 +1,32 @@
...
+
+	// Link to add-on.
+	link Addon AddOn
```
jhernand, line 21: I think we should start to avoid these "back links". Then aren't really useful and they introduce circular references that are harmless now but can block evolution of the architecture in the future. So please remove it.

model/clusters_mgmt/v1/add_on_type.model
```diff
@@ -52,10 +52,12 @@ class AddOn {
...
 
-	// Indicates if this add-on has external resources associated with it 
+	// Indicates if this add-on has external resources associated with it
 	HasExternalResources Boolean
 
 	// List of parameters for this add-on.
 	link Parameters []AddOnParameter
 
+	// List of requirements for this add-on.
+	link Requirements []AddOnRequirement
```
jhernand, line 62: The `link` is intended for when the actual collection is in another enpoint, but there is no such endpoint, or at least I can't find it.  The same applies to the existing `Parameters` attribute above.

model/clusters_mgmt/v1/add_on_requirement_type.model
mikenairn: Its' a struct, will update

mikenairn:
Thanks for the feedback @jhernand, was helpful. Does this look better now? I'll deal with the issues with parameters in a separate PR

jhernand (approved):
Looks good to me, thanks @mikenairn.

jhernand:
@mikenairn remove the `WIP` and I will merge.

PR #313, opened 2021-01-28
Add cluster hibernation support

nimrodshn:
cc @igoihman

### DIFF
--- a/model/clusters_mgmt/v1/cluster_resource.model
+++ b/model/clusters_mgmt/v1/cluster_resource.model
@@ -75,6 +75,17 @@ resource Cluster {
 		target MachinePools
 	}
 
+	// Initiates cluster hibernation. While hibernating a cluster will not consume any cloud provider infrastructure
+	// but will be counted for quota.
+    method Hibernate {
+
+    }
+
+    // Resumes from Hibernation.
+    method Resume {
+
+    }
+
 	// Reference to the resource that manages metrics queries for the cluster.
 	locator MetricQueries {
 		target MetricQueries


nimrodshn:
@igoihman Please note that the current API is not supported using the ocm-model / sdk and @jhernand  suggested that we change our "resume from hibernation" / "unhibernate API from :`DELETE /api/clusters_mgmt/v1/clusters/<id>/actions/hibernate` to `POST /api/clusters_mgmt/v1/clusters/<id>/actions/unhibernate`

@jhernand  had also suggested we drop the `/actions` preface, 

@tzvatot @igoihman WDYT?

igoihman:
@nimrodshn let's drop the `/actions` preface. The suggested approach sounds good, what about calling the new endpoint `resume`?

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -75,6 +75,17 @@ resource Cluster {
 		target MachinePools
 	}
 
+	// Initiates cluster hibernation; While hibernating a cluster will not consume any cloud provider infrastructure
+	// but will be counted for quota.
```
jhernand, line 79: This documentation isn't currently used for anything other than putting it in the OpenAPI documentation. But in the future I would like to use it to generate human readable documentation. In order to do that I would like to have in each piece of documentation a short sentence describing the thing. Then a dot, and then possible additional, longer, documentation. That dot will be used by the tool that generates the human readable documentation. So, can you reformat like this?

```
// Initiates cluster hibernation. While hibernating a cluster will not consume any cloud provider infrastructure
// but will be counted for quota.
```

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -75,6 +75,17 @@ resource Cluster {
...
+    method Hibernate {
```
jhernand, line 80: Make sure to use tabs instead of spaces for indentation.

jhernand (commented):
Looks good. Just some minor formatting comments. Address them an I will merge.

model/clusters_mgmt/v1/cluster_resource.model
nimrodshn: Done. thanks

model/clusters_mgmt/v1/cluster_resource.model
nimrodshn: Thanks, done.

jhernand (approved):
Looks good, thanks.

PR #314, opened 2021-01-29
WIP Update addon parameter models

mikenairn:
Change addon parameter type from class to struct
Remove unnecessary links from addon parameter list types

### DIFF
--- a/model/clusters_mgmt/v1/add_on_installation_type.model
+++ b/model/clusters_mgmt/v1/add_on_installation_type.model
@@ -32,7 +32,7 @@ class AddOnInstallation {
 	OperatorVersion String
 
 	// List of add-on parameters for this add-on installation.
-	link Parameters []AddOnInstallationParameter
+	Parameters []AddOnInstallationParameter
 
 	// Date and time when the add-on was initially installed in the cluster.
 	CreationTimestamp Date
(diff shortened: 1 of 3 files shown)


model/clusters_mgmt/v1/add_on_installation_type.model
```diff
@@ -32,7 +32,7 @@ class AddOnInstallation {
 	OperatorVersion String
 
 	// List of add-on parameters for this add-on installation.
-	link Parameters []AddOnInstallationParameter
+	Parameters []AddOnInstallationParameter
```
jhernand, line 35: I think that this is correct. But it also has an important side effect in the JSON repreentation. With this "link" the representation looked like this:

```
"parameters": {
  "items": {
    {
      "id": "my-parameter",
      "name": "My parameter",
      ...
    },
    {
      "id": "my-parameter",
      "name": "My parameter",
      ...
    },
   ...
  }
```

After changing this the JSON representation will be:

```
"parameters": [
  {
    "id": "my-parameter",
    "name": "My parameter",
    ...
  },
  {
    "id": "my-parameter",
    "name": "My parameter",
    ...
  },
]
```

Note how `parameters` is now an array instead of an object, and how the `items` intermediate array has been removed. The reason for that is that those are only needed for links to collections, but not for regular arrays. This means that you will have to change the server accordingly.

mikenairn: @jhernand Thanks, i made the required updates. This is actually how i originally wanted it to be, but i mistakenly added the "link" when i had no need to which forced me down the wrong path.

PR #317, opened 2021-02-08
Add billing_model field to cluster type

zgalor:
cc @rawsyntax 
@tzvatot @igoihman @jhernand PTAL

### DIFF
--- a/model/clusters_mgmt/v1/billing_model_type.model
+++ b/model/clusters_mgmt/v1/billing_model_type.model
@@ -1,0 +1,24 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Billing model for cluster resources.
+enum BillingModel {
+	// Standard. This is the default billing model
+	Standard
+
+	// Red Hat Marketplace billing model.
+	Marketplace
+}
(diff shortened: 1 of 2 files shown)


nimrodshn:
LGTM :100:

model/clusters_mgmt/v1/cluster_type.model
```diff
@@ -190,4 +190,7 @@ class Cluster {
 
 	// Channel group to be used for upgrading the cluster. Valid values: stable (default), fast, candidate, nightly.
 	UpgradeChannelGroup String
+
+	// Billing model for cluster resources. Valid values: standard (default), marketplace.
+	BillingModel String
```
jhernand, line 195: Is there any reason to not use an _enum_ type for this?

zgalor: No. I'll fix it

zgalor:
@jhernand rePTAL

jhernand (approved):
Looks good to me. Thanks @zgalor.

model/clusters_mgmt/v1/billing_model_type.model
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+// Billing model for cluster resources.
+enum BillingModel {
+	// Standard. This is the default billing model
+	Standard
+
+	// Red Hat Marketplace billing model.
+	Marketplace
```
petli-openshift, line 23: both should be lower-case. sry for the misleading instructions in the card. We've changed it after the initial design.

@rawsyntax I think we should update our code to be case-insensitive. Now it rejects `Standard`,
```
{
  "id": "8",
  "kind": "Error",
  "href": "/api/accounts_mgmt/v1/errors/8",
  "code": "ACCT-MGMT-8",
  "reason": "BillingModel is invalid: Standard",
  "operation_id": "1oCq468gjV68YkeVoknLybqyjSm"
}
```

zgalor: @petli-openshift ocm api only supports lower case inputs.
https://github.com/openshift-online/ocm-api-metamodel project generates the SDK and the openapi yaml file and translates these camel case values to snake case.

PR #323, opened 2021-02-16
add billing_model to reserved_resource_type

petli-openshift:


petli-openshift:
this mr adds billing_model to,
- ReservedResource: it will be used by cluster_authorization request and can be accessed by reserved_resource
- SubscriptionResource: it can be accessed by subscription

jhernand (approved):
Looks good to me.

PR #332, opened 2021-03-15
Add 'deprovision' and 'force' parameters to delete cluster method

zgalor:


zgalor:
@jhernand @igoihman @tzvatot @nimrodshn PTAL

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -28,6 +28,14 @@ resource Cluster {
 
 	// Deletes the cluster.
 	method Delete {
+		// true by default. If false it will only delete from OCM but not the actual cluster resources.
```
jhernand, line 31: I plan to use the first sentence of the documentation as the summary in generated documentation files. In this case the summary for `deprovision` will be `true by default`. That isn't very informative. Can you switch the order of these sentences so that `true by default` is the last one?

Also note that the model and metamodel support explicitly specifying default values of parameters, so you can do this:

```
in Deprovision Boolean = true
```

That will have no impact in the generated client code. But it will in the generated server code (which we are not using yet) as it will automatically set the value of the parameter to the specified default when not explicitly set by the caller.

Also try to use camel case for parameter names: `Deprovision` in this case. This is for consistency with other parameters in other methods. We should probably change the model and metamodel so that the names here are the sames that are actually used in the HTTP messages, but for now I think that is better to stay consistent with what we have.

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -28,6 +28,14 @@ resource Cluster {
...
+		// false is only allowed for OCP clusters.
+		in deprovision Boolean
+
+		// false by default. If true it will force delete the cluster even if external dependecies
+		// cleanup was not finalized. Should be used with extereme caution, after manual cleanup
+		// of external dependencies. OSD only.
+		in force Boolean
```
jhernand, line 38: Same.

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -28,6 +28,14 @@ resource Cluster {
 
 	// Deletes the cluster.
 	method Delete {
+		// true by default. If false it will only delete from OCM but not the actual cluster resources.
+		// false is only allowed for OCP clusters.
+		in deprovision Boolean
+
+		// false by default. If true it will force delete the cluster even if external dependecies
+		// cleanup was not finalized. Should be used with extereme caution, after manual cleanup
```
jhernand, line 36: s/extereme/extreme/

jhernand (changes requested):

zgalor:
@jhernand I fixed the issues. PTAL

jhernand (approved):
Looks good to me. Thanks @zgalor.

PR #333, opened 2021-03-16
SDA-3849 - fix addon installation in model

ciaranRoche:
Closes : https://issues.redhat.com/browse/SDA-3849

Currently we have the wrong object for an AddOn Installation. I created a gist to show the example as how it is outlined in the api [1] versus what the object should look like.  

@jhernand @mikenairn @vkareh PTAL 
Currently the addOnInstallation is a direct representation of the object within Cluster Service. Does this need to be declared in the API, or can we just declare the link to the `AddOn` and the `AddOnInstallationParameters`

[1] - https://api.openshift.com/#/default/post_api_clusters_mgmt_v1_clusters__cluster_id__addons

### DIFF
--- a/model/clusters_mgmt/v1/add_on_installation_type.model
+++ b/model/clusters_mgmt/v1/add_on_installation_type.model
@@ -16,27 +16,9 @@ limitations under the License.
 
 // Representation of an add-on installation in a cluster.
 class AddOnInstallation {
-	// ID used to identify the cluster that this add-on is attached to.
-	link Cluster Cluster
-
 	// Link to add-on attached to this cluster.
 	link Addon AddOn
 
-	// Overall state of the add-on installation.
-	State AddOnInstallationState
-
-	// Reason for the current State.
-	StateDescription String
-
-	// Version of the operator installed by the add-on.
-	OperatorVersion String
-
 	// List of add-on parameters for this add-on installation.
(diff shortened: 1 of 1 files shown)


jhernand:
@ciaranRoche I don't see the gist that you mention. Can you share it?

mikenairn:
@ciaranRoche Does this fix the problem? I'm not sure how this makes any difference, is it not showing the AddOn model in the AddOnInstallation example response rather than this model in the swagger docs?

mikenairn:
Ah nvm, i see what it's doing, it's including the addon and the cluster, where we don't actually return those entire objects from the API, only the the href

ciaranRoche:
Sorry guys @jhernand @mikenairn , i forgot to link the gist https://gist.github.com/ciaranRoche/4577cc82b13bbf1e37f94762596aa7cf

jhernand:
The "current_example" that you put in that gist is automatically generated by the Swagger tool, we don't create it explicitly. That tool has no way of knowing that in certain cases you will include attributes in your responses and in some cases you wont'. I think it is not correct to remove the attributes from the API specification (and therefore from the SDK) just because we don't want to have those incorrect examples. If we wan't better examples then I'd suggest to add them to the documentation of the model. There we can clearly explain what will be returned, and when the attributes will be populated and when will be empty.

PR #336, opened 2021-03-29
aws: Support PrivateLink for fully-private clusters

vkareh:


model/clusters_mgmt/v1/aws_type.model
```diff
@@ -27,4 +27,7 @@ struct AWS {
 
 	// SubnetIDs are the subnet ids to be used when installing the cluster.
 	SubnetIDs []String
+
+	// PrivateLink sets cluster to be inaccessible externally
```
jhernand, line 31: There is no need to repeat the name of the attribute in the first word of the comment because this isn't going to be processed by the Go documentation tool. Actually that will result in less than ideal documentation, so just remove tht word. Also end the sentence with a dot. Once that is done feel free to merge.

PR #345, opened 2021-04-13
Add event and site code to Terms and SelfTerms requests

bvulaj:
These will default to `ocm` and `register` if not provided. This was the previous behavior.


tiwillia:
LGTM

PR #366, opened 2021-05-13
Updated Arguments to JobQueue#Push

gshilin-sdb:


model/job_queue/v1/queue_resource.model
```diff
@@ -27,10 +27,10 @@ resource Queue {
         out Kind String
         out HREF String
         in out Arguments String
-        out Attempts Integer
-        out AbandonedAt Date
+        in out Attempts Integer
+        in out AbandonedAt Date
         out ReceiptId String
-        out CreatedAt Date
+        in out CreatedAt Date
```
jhernand, line 33: So the user that pushes a new job can decide what is the creation date of that job?

model/job_queue/v1/queue_resource.model
```diff
@@ -27,10 +27,10 @@ resource Queue {
         out Kind String
         out HREF String
         in out Arguments String
-        out Attempts Integer
-        out AbandonedAt Date
+        in out Attempts Integer
```
jhernand, line 30: What does this mean? That the user pushing a job decides the maximum number of attempts to execute the job? Would be nice to have that explained in the documentation.

model/job_queue/v1/queue_resource.model
```diff
@@ -27,10 +27,10 @@ resource Queue {
...
+        in out AbandonedAt Date
```
jhernand, line 31: Same. Without additional documentation I would probably assume that this means that the job will be automatically abandoned at the date that the user specifies here. Is that correct?

model/job_queue/v1/queue_resource.model
gshilin-sdb: Talked yesterday to Brandon. Yes, you can re-push a job and to keep its creation time

model/job_queue/v1/queue_resource.model
gshilin-sdb: AWS SQS permits to supply every field. None of them have special meaning. Both Attempts and AbandonedAt can be supplied at creation time.

gshilin-sdb: And technically we don't care what exactly client wants to specify

gshilin-sdb: These fields can be specified when you want to push the same job again to try more times. We're not advising for it, but we don't want to hide this as well.

gshilin-sdb: We don't want it to be in documentation. From end-user's point of view only `Arguments` are relevant at Push time

jhernand: I find it hard to understand this approach. I though we were building a service to manage jobs, and that SQS was just an implementation detail. But from your words it looks like the SQS details leak through the API of the service.

Anyhow, I assume you have considered this already, and decided it is a good way to proceed. I won't get in the way.

jhernand (approved):

PR #376, opened 2021-05-31
Remove Dashboards resource

nimrodshn:
This resource was recently removed from `clusters-service`.

cc: @igoihman @tzvatot @jhernand


jhernand (approved):
Looks good to me.

PR #379, opened 2021-06-01
sts: Add support role ARN

vkareh:
To enable SREs to access an STS cluster's AWS account in order to
provide support, we expose a support role ARN attribute that gets set in
the accountclaim.


vkareh:
@pvasant - please take a look

related: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2990

pvasant:
LGTM

PR #392, opened 2021-07-06
Add username to service log entry

zgalor:
@tzvatot @nimrodshn PTAL

### DIFF
--- a/model/service_logs/v1/cluster_log_type.model
+++ b/model/service_logs/v1/cluster_log_type.model
@@ -42,4 +42,7 @@ class LogEntry {
 	InternalOnly Boolean
 
 	Timestamp Date
+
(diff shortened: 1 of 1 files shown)


model/service_logs/v1/cluster_log_type.model
```diff
@@ -42,4 +42,7 @@ class LogEntry {
 	InternalOnly Boolean
 
 	Timestamp Date
+
+	// The username that triggered the event (unless internal).
```
tzvatot, line 46: I would change the description "unless internal" to "if available"

zgalor: fixed

tzvatot:
LGTM

PR #400, opened 2021-08-03
Added kms key arn to model

pvasant:


pvasant:
@vkareh @tzvatot  PTAL

vkareh:
LGTM

PR #407, opened 2021-08-23
[SDB-2229] Added support for OSL /clusters/<uuid>

Lir10:


tzvatot:
LGTM

model/service_logs/v1/cluster_uuid_resource.model
```diff
@@ -0,0 +1,24 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a specific Cluster UUID.
+resource ClusterUUID {
```
jhernand, line 18: Is there any reason to not name this jut `Cluster`?

Lir10: Sure, I've changed it

model/service_logs/v1/clusters_resource.model
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the collection of clusters for clusters logs.
+resource Clusters {
+
+	// Reference to the service that manages a specific ClusterUUID.
+	locator ClusterUUID {
```
jhernand, line 21: Should this also be named `Cluster`?

Lir10: done

model/service_logs/v1/cluster_resource.model
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a specific Cluster for a cluster log.
+resource Cluster {
+
+	// Reference to the list of cluster logs for a specific cluster uuid.
+	locator ClusterLogs {
```
jhernand, line 21: This should probably be just `Log`, that way the code to access the logs will look like this:

```go
logsResource := connection.ServiceLogs().V1().Clusters().Cluster("123").Logs()
```

Otherwise it would be like this, which repeats the "cluster" word many times:

```go
logsResource := connection.ServiceLogs().V1().Clusters().Cluster("123").ClusterLogs()
```

Lir10: @jhernand We can change that in the SDK but that won't be consistent with our original path `/api/service_logs/v1/clusters/{uuid}/cluster_logs`

jhernand: OK, don't change it.

Lir10: Note that `cluster_logs` is on purpose because most the paths in OSL end with `cluster_logs` so we add that for consistency. 

For example to list/post logs the path looks like this `/api/service_logs/v1/cluster_logs` so that's why we chose `/api/service_logs/v1/clusters/{uuid}/cluster_logs`

jhernand: I understand, thanks for explaining.

PR #415, opened 2021-09-09
Update changes doc

zgalor:
@tzvatot PTAL


zgalor:
merging this, @tzvatot FYI

PR #417, opened 2021-09-13
added archived cluster endpoint

pvasant:


model/clusters_mgmt/v1/archived_clusters_resource.model
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright (c) 2019 Red Hat, Inc.
```
vkareh, line 2: 2021

model/clusters_mgmt/v1/root_resource.model
```diff
@@ -76,10 +76,14 @@ resource Root {
...
-	
+
 	// Reference to limited support reason templates.
 	locator LimitedSupportReasonTemplates {
 	    target LimitedSupportReasonTemplates
 	}
 
+	locator ArchivedClusters {
+	    target ArchivedClusters
+    }
```
vkareh, line 87: The indentation here looks a bit off

vkareh:
Other than an indentation issue in the root resource, this LGTM

pvasant:
@vkareh  ah i dont know what is the issue with vscode or goland. I use only tabs not spaces still indentation is always off and i had to edit directly to fix it. not sure which editor is the right one to avoid this

PR #424, opened 2021-10-04
move additional_trust_bundle from cluster to the Proxy object

zvikorn:


vkareh:
LGTM

Make sure that the API handler can support both versions until all known clients have upgraded the SDK.

PR #436, opened 2021-11-18
SDA 4786 - Add Channel to AddOnVersion

ciaranRoche:
Part of : https://issues.redhat.com/browse/SDA-4786

SDK Build Verification - https://gist.github.com/ciaranRoche/c9225aa4f9169c6700774ce71fa4e1d3

### DIFF
--- a/model/clusters_mgmt/v1/add_on_install_mode_type.model
+++ b/model/clusters_mgmt/v1/add_on_install_mode_type.model
@@ -18,7 +18,7 @@ limitations under the License.
 enum AddOnInstallMode {
 	// This mode means that the the addon CRD exists in a single specific namespace.
 	// This namespace is reflected by the TargetNamespace addon field
-	SingleNamespace
+	OwnNamespace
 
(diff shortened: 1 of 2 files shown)


ciaranRoche:
@mikenairn PTAL

model/clusters_mgmt/v1/add_on_install_mode_type.model
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 enum AddOnInstallMode {
 	// This mode means that the the addon CRD exists in a single specific namespace.
 	// This namespace is reflected by the TargetNamespace addon field
-	SingleNamespace
+	OwnNamespace
```
mikenairn, line 21: Is the comment still valid, or does that need updating?

ciaranRoche: From my understanding its essentially the same thing, just with the new Sync Set change, MT only support Own and All namespaces. Own Namespace come from what is defined in target namespace

PR #445, opened 2021-12-06
Add Status Board models

djberg96:
This PR adds the initial resources/models for the the Status Board application.

### DIFF
--- a/model/status_board/v1/application_dependency_type.model
+++ b/model/status_board/v1/application_dependency_type.model
@@ -1,0 +1,42 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Definition of a Status Board application dependency.
+class ApplicationDependency {
+  // Object creation timestamp.
+  CreatedAt Date
+
+  // Object modification timestamp.
+  UpdatedAt Date
+
+  // The name of the application. 
+  Name String
+
+  // Miscellaneous metadata about the application.
+  Metadata Interface
+
+  // The type of application dependency, e.g. soft or hard.
+  Type String
+
+  // The Service associated with the dependency.
+  Service Service
+
+  // The parent Application of the dependency.
+  Application Application
+
+  // The application dependency owners (name and email)
+  Owners []Owner
+}
--- a/model/status_board/v1/applications_resource.model
+++ b/model/status_board/v1/applications_resource.model
@@ -1,0 +1,37 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the collection of applications.
+resource Applications {
+  // Retrieves the list of applications.
+  method List {
+    in Fullname String
+    in OrderBy String
+    in out Page Integer = 1
+    in out Size Integer = 100
+    out Total Integer
+    out Items []Application
+  }
+
+  method Add {
+    in out Body Application
+  }
+
+  locator Application {
+    target Application
+    variable ID
+  }
+}
--- a/model/status_board/v1/owner_type.model
+++ b/model/status_board/v1/owner_type.model
@@ -1,0 +1,24 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Definition of a Status Board owner.
+class Owner {
+  // The owner's username.
+  Username String
+
+  // The owner's email address.
+  Email String
+}
--- a/model/status_board/v1/peer_dependencies_resource.model
+++ b/model/status_board/v1/peer_dependencies_resource.model
@@ -1,0 +1,36 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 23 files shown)


model/status_board/v1/application_dependency_type.model
```diff
@@ -0,0 +1,42 @@
...
+  Name String
+
+  // Miscellaneous metadata about the application.
+  Metadata Interface
+
+  // The type of application dependency, e.g. soft or hard.
+  Type String
+
+  // The Service associated with the dependency.
+  ServiceId String
```
jhernand, line 38: In general it is better to have references to other objects implemented using the types corresponding to those objects. Here you could do this:

```
// The Service associated with the dependency:
Service Service
```

The JSON representation of that would then be something like this:

```json
{
  "kind": "ApplicationDependency",
  "id": "123",
  "href": "/api/status_board/v1/application_dependencies/123",
  "service": {
    "kind": "ServiceLink",
    "id": "456",
    "href": "/api/status_board/v1/services/456"
  },
  ...
}
```

Note how the `kind` attribute is `ServiceLink` instead of `Service`. That is what indicates that this is a reference to an object and not the object itself. Tools can then use this information to automatically retrieve the complete object if needed by just taking the value of the `href`  attribute and sending a `GET` request.

This is just a recommendation. You will find existing services that follow it and some that don't.

Same for the reference to the application below.

model/status_board/v1/applications_resource.model
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the collection of applications.
+resource Applications {
+  // Retrieves the list of applications.
+  method List {
+    in Fullname String
```
jhernand, line 21: What is the meaning of this `fullname` parameter? May be worth explaining it in the documentation.

model/status_board/v1/applications_resource.model
```diff
@@ -0,0 +1,37 @@
...
+    in OrderBy String
```
jhernand, line 22: The name of this parameter should be `Order`.

model/status_board/v1/owner_type.model
```diff
@@ -0,0 +1,33 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Definition of a Status Board owner.
+class Owner {
```
jhernand, line 18: Does this type have its own `id`? If it doesn't then it should probably be a `struct` instead of a `class`. The difference between struts and classes is that classes are expected to have their own identity, and therefore the metamodel tool will automatically add the `id` attribute.

model/status_board/v1/peer_dependencies_resource.model
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License.
+*/
+
+// Manages the collection of peer dependencies.
+resource PeerDependencies {
+  // Retrieves the list of peer dependencies.
+  method List {
+    in out Page Integer = 1
+    in out Size Integer = 100
+    in OrderBy String
```
jhernand, line 23: Should be `order`.

model/status_board/v1/peer_dependency_resource.model
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Provides detailed information about the specified peer dependency.
+resource PeerDependency  {
+  method Get {
+    out Body Service
```
jhernand, line 20: I think this `Body` parameter should be of type `PeerDependency`.

model/status_board/v1/products_resource.model
```diff
@@ -0,0 +1,37 @@
...
+limitations under the License.
+*/
+
+// Manages the collection of products.
+resource Products {
+  // Retrieves the list of products.
+  method List {
+    in out Page Integer = 1
+    in out Size Integer = 100
+    in Fullname String
```
jhernand, line 23: What is the meaning of this `fullname` parameter?

model/status_board/v1/products_resource.model
```diff
@@ -0,0 +1,37 @@
...
+    in OrderBy String
```
jhernand, line 24: This should be `Order`.

model/status_board/v1/service_type.model
```diff
@@ -0,0 +1,55 @@
...
+  // Object deletion timestamp.
+  DeletedAt Date
+
+  // The Application associated with the Service
+  ApplicationId String
+
+  // The name of the Service
+  Name String
+
+  ServiceEndpoint String
```
jhernand, line 34: What is the meaning of this `service_endpoint` attribute? Would be nice to have this (and the other attributes) documented.

model/status_board/v1/service_type.model
```diff
@@ -0,0 +1,55 @@
...
+
+  StatusType String
```
jhernand, line 36: This looks like an enumerated type. What are the possible values? Consider using an `enum` to better document it.

model/status_board/v1/services_resource.model
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+// Manages the collection of services.
+resource Services {
+  // Retrieves the list of services.
+  method List {
+    in out Page Integer = 1
+    in out Size Integer = 100
+    in OrderBy String
+    in Fullname String
```
jhernand, line 24: What is the meaning of `fullname`?

model/status_board/v1/services_resource.model
```diff
@@ -0,0 +1,38 @@
...
+    in Mine Boolean
```
jhernand, line 25: What is the meaning of `mine`? I guess it is to filter the results so the user only gets the services that she owns. If that is correct then consider using `search` for that.

model/status_board/v1/status_type.model
```diff
@@ -0,0 +1,36 @@
...
+  CreatedAt Date
+
+  // Object modification timestamp.
+  UpdatedAt Date
+
+  // Object deletion timestamp.
+  DeletedAt Date
+
+  // The associated service ID.
+  ServiceId String
```
jhernand, line 29: Consider using a link.

model/status_board/v1/statuses_resource.model
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+// Manages the collection of statuses
+resource Statuses {
+  // Retrieves the list of statuses.
+  method List {
+    in out Page Integer = 1
+    in out Size Integer = 100
+    in CreatedAfter Date
+    in CreatedBefore Date
```
jhernand, line 24: Consider using `search` instead of these custom search criteria: `search=created_at >= ... and created_at <= ...`.

jhernand (commented):
Most of my comments are just recommendations. Only a couple of them are potential errors.

I see that you aren't adding the `search` parameter to list operations. Consider using it instead of custom search criteria.

model/status_board/v1/owner_type.model
djberg96: Yes, it has its own ID.

model/status_board/v1/peer_dependencies_resource.model
djberg96: The `orderBy` is a query param that we've already defined in our own openapi.yaml config.

model/status_board/v1/service_type.model
djberg96: It's a text field in our schema, it's not normalized.

djberg96:
@jhernand Regarding the `search` parameter that you mentioned in a few places, we don't allow it due to fear of sql injection.

[I will replace this with an internal link to the discussion if I'm allowed, checking]

In essence, I am "backporting" our allowed query params from an existing openapi.yaml config.

model/status_board/v1/application_dependency_type.model
djberg96: Ok, updated.

model/status_board/v1/status_type.model
djberg96: Done!

jhernand (commented):
Looks good to me. Please squash to a single commit before merging.

djberg96:
@jhernand Squashed.

jhernand (approved):

PR #452, opened 2021-12-16
Add statuses path to service model

djberg96:
This adds a missing statuses endpoint to the services path, plus a few comments.

### DIFF
--- a/model/status_board/v1/product_resource.model
+++ b/model/status_board/v1/product_resource.model
@@ -27,10 +27,12 @@ resource Product {
   method Delete {
   }
 
+  // /api/status-board/v1/products/{id}/applications
   locator Applications {
     target Applications
   }
 
+  // /api/status-board/v1/products/{id}/updates
   locator Updates {
     target Statuses
   }
--- a/model/status_board/v1/product_resource.model
+++ b/model/status_board/v1/product_resource.model
@@ -27,12 +27,10 @@ resource Product {
   method Delete {
   }
 
-  // /api/status-board/v1/products/{id}/applications
   locator Applications {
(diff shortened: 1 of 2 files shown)


model/status_board/v1/product_resource.model
```diff
@@ -27,10 +27,12 @@ resource Product {
   method Delete {
   }
 
+  // /api/status-board/v1/products/{id}/applications
```
jhernand, line 30: I think it is redundant to have this here, as this is automatically calculated by the model and added to the generated OpenAPI documentation. If you really want to include it then add a first sentence explaining that this is, and make this a complete example:

```
// Reference to the service that manages the collection
// of applications associated to this product. For example
// go get the list of applications for product `123` send a
// request like this:
//
// ```http
// GET /api/status-board/v1/products/123/applications
// ```
```

model/status_board/v1/product_resource.model
```diff
@@ -27,10 +27,12 @@ resource Product {
...
   locator Applications {
     target Applications
   }
 
+  // /api/status-board/v1/products/{id}/updates
```
jhernand, line 35: Same.

model/status_board/v1/service_resource.model
```diff
@@ -26,4 +26,9 @@ resource Service  {
 
   method Delete {
   }
+
+  // /api/status-board/v1/services/{id}/statuses
```
jhernand, line 30: Same.

djberg96:
@jhernand Ok, updated.

jhernand (approved):

PR #454, opened 2021-12-19
[SDB-2509] Update OSL API schema to be compatible with ocm-sdk-go

Lir10:


model/service_logs/v1/cluster_log_type.model
```diff
@@ -45,4 +45,12 @@ class LogEntry {
...
 	// The username that triggered the event (if available).
 	Username String
+	
+	// Internal cluster ID.
+	ClusterID String
+
+	// Log custom event id for a simple search of related cluster logs.
+	EventStreamID String
+
+	SubscriptionID String
```
tzvatot, line 55: Would you like to add a comment for this one?

Lir10: Added - thanks

PR #458, opened 2021-12-20
Change version gates resource to be gates instead of version_gates

tzvatot:
Change locator name to make sure the URL will be /versions/<version-id>/gates (instead of "version_gates").


tzvatot:
@zgalor @zvikorn PTAL

zgalor:
lgtm

PR #473, opened 2022-01-10
Release v0.0.168

vkareh:
- Fix description of various API attributes
- OVN: Add network type selection

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.0.168 Jan 10 2022
+- Fix description of various API attributes
+- OVN: Add network type selection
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.0.168 Jan 10 2022
+- Fix description of various API attributes
```
tzvatot, line 6: what about "WIP: adding field to hold validation error message" or you're not including it on purpose?

ciaranRoche: I think that was merged after this MR was created, be awesome to include it though :pray:

vkareh: rebased

PR #474, opened 2022-01-12
Add ServiceInfo to status board Status model

PanSpagetka:
Add ServiceInfo to status board Status model.

### DIFF
--- a/model/status_board/v1/service_info_type.model
+++ b/model/status_board/v1/service_info_type.model
@@ -1,0 +1,24 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Definition of a Status Board service info.
+struct ServiceInfo {
+  // Full name of the service
+  Fullname string
+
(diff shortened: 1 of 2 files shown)


model/status_board/v1/status_type.model
```diff
@@ -25,6 +25,9 @@ class Status {
   // The associated service ID.
   Service Service
 
+  // Additional Service related data.
+  ServiceInfo Interface
```
vkareh, line 29: Why is this an interface? Why not create a named struct with the necessary attributes in it?
```
class ServiceInfo {
  Fullname string
  StatusType string
}
```

vkareh:
Since the API uses a `map[string]interface{}` type, you are better off setting the model type as `[String]Interface`, as it represents the type better. However, I'm curious about this approach.

PanSpagetka:
We want to treat this as a JSON, but when I think about it, there is no reason this couldn't be a new type. I made a new MR to create new `ServiceInfo` type and also changed this PR. Thank you for suggestion :)

djberg96:
@PanSpagetka Looks like the service_info_type.model is missing. It would look like @vkareh's comment, i.e.

```
class ServiceInfo {
  Fullname String
  StatusType String
}
```

model/status_board/v1/service_info_type.model
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+// Definition of a Status Board service info.
+class ServiceInfo {
+  // Full name of the service
+  Fullname string
+
+  // Type of the service status
+  StatusType string
+}
```
jhernand, line 24: Using a `class` here instead of a `struct` means that this is anentity has its own identity, and in practical terms that it will have the `id` and `href` attributes. With my limited understanding of your service I'd say that is not what you want or need. So better replace `class` with `struct`.

PanSpagetka: Thank you, `class` replaced with `struct`.

jhernand (approved):
Looks good to me, thanks @PanSpagetka.

PR #493, opened 2022-02-17
add list of requirements to addon parameter options

jmolmo:
JIRA - sda-5400

Signed-off-by: Juan Miguel Olmo Martínez <jolmomar@redhat.com>


ciaranRoche:
small NP, can you update the commit message to `add list of requirements to addon parameter options` just to make it clearer for whoever is makes a release.

PR #495, opened 2022-02-23
Adding the Managed Service api

renan-campos:


model/service_mgmt/v1/root_resource.model
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Root of the tree of resources of the Management Services.
+resource Root {
+	// Reference to the resource that manages the collection of Managed Services.
+	locator Service {
```
jhernand, line 20: This is a collection, should be `services`. Are you sure you want URLs to be `/api/service_mmt/v1/service/123...`?

model/service_mgmt/v1/root_resource.model
```diff
@@ -0,0 +1,23 @@
...
+		target Service
```
jhernand, line 21: The target resource should also probably be named `Services`.

model/service_mgmt/v1/service_resource.model
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages Managed Services.
+resource Service {
+
+  // Creates a new Managed Service
+	method Add {
```
jhernand, line 21: Sounds strange that you have a method to add a service, but no methods to list services or get a specific service. Is that on purpose?

model/service_mgmt/v1/service_type.model
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Definition of a Managed Service Request
+class Service {
+	// Name of the service. This will match the name of the add-on that
+	// the Managed Service is to deploy.
+	Service String
```
jhernand, line 21: If this is the name why not name it `name`?

model/service_mgmt/v1/cluster_parameters_type.model
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This represents the parameters needed by Managed Service to create a cluster.
+struct ClusterParameters {
+	Name String
+	DisplayName String
+	link Region CloudRegion
```
jhernand, line 21: If this is a link where is in the target of that link? I don't see in this API definition any method that returns those regions. Is this intended to be a link to some other service?

model/service_mgmt/v1/cluster_parameters_type.model
```diff
@@ -0,0 +1,24 @@
...
+	link AWS AWS
```
jhernand, line 22: Same, where is the method that returns this AWS objects?

PR #500, opened 2022-03-02
Add 'environment' endpoint - get & patch

olisakov:
added environment endpoint - get &patch
updated the changes to the last version
@etabak

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 This document describes the relevant changes between releases of the API model.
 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.1.175 Mar 02 2022
+- Add environment endpoints get and patch
```
tzvatot, line 6: It's only a single endpoint, so please use singular instead of plural.

PR #501, opened 2022-03-03
Fix service_logs API exposes unimplemented POST method for /api/servi…

Lir10:


Lir10:
@jhernand ptal

model/service_logs/v1/cluster_logs_uuid_resource.model
```diff
@@ -0,0 +1,69 @@
...
+
+		// Retrieved list of Cluster logs.
+		out Items []LogEntry
+	}
+
+    // Reference to the service that manages a specific Log entry.
+	locator LogEntry {
+		target LogEntry
+		variable ID
+	}
```
jhernand, line 68: I this really supported? I think there are no sub-resources of this resource.

Lir10: LogEntry is a class in `cluster_log_type.model`

jhernand: This locator translates into a `/api/service_logs/v1/clusters/{cluster_id}/cluster_logs/{entry_id}` endpoint. I think we don't have such an endpoint in the server, do we? If we don't have that endpoint then this locator can and should be removed.

Lir10: I see .. ok I'll remove that one. thanks

PR #510, opened 2022-03-09
Fix path segment of cluster logs

jhernand:
The URL of the path segment for the logs of a specific cluster should be
`/api/service_logs/clusters/{id}/cluster_logs` but due to an error in
the corresponding locator the last segment is currently
`cluster_logs_uuid`. This patch fixes that.


jhernand:
FYI: @Lir10 @zvikorn

PR #517, opened 2022-03-14
Add ReleaseImage to Version

samira-barouti:
In order to spin up Hypershift clusters, we need to be able to retrieve the
release image info from the Version object. This PR aims to add
this info by adding a String field to Version model.

* Relates SDA-5531


jhernand (approved):
Looks good, thanks @samira-barouti .

PR #525, opened 2022-03-30
Add ManagementCluster to ProvisionShard

samira-barouti:
* Relates SDA-5524

### DIFF
--- a/model/clusters_mgmt/v1/provision_shard_type.model
+++ b/model/clusters_mgmt/v1/provision_shard_type.model
@@ -37,4 +37,8 @@ class ProvisionShard {
 
 	// Contains the cloud provider name
 	link CloudProvider CloudProvider
+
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/provision_shard_type.model
```diff
@@ -37,4 +37,7 @@ class ProvisionShard {
 
 	// Contains the cloud provider name
 	link CloudProvider CloudProvider
+
+	// Contains the name of Managment Cluster for Hypershift clusters
```
jhernand, line 41: Would be nice if you can explain in the documentation that is the currently selected management cluster for this shard, that it will change over time, and that t is read-only, i.e. it can't be changed via the API.

samira-barouti: Done.

jhernand (approved):
Looks good to me, thanks @samira-barouti .

PR #531, opened 2022-04-01
Add Status query param for incidents resource

djberg96:
After making this change I ran `>make model_version=HEAD model_url=/Users/dberger/Dev/ocm-api-model generate`. That seemed to work.

However, if I then try to build an example web-rca program with those changes I get the following:

```
>go build examples/list_webrca_incidents.go

# github.com/openshift-online/ocm-sdk-go/webrca/v1
webrca/v1/incidents_resource_json.go:102:9: request.status undefined (type *IncidentsListServerRequest has no field or method status)
```

It looks like it generated a `Status_` function and `status_` fields instead for some reason. Is the word "status" a keyword of some sort? What am I doing wrong?

### DIFF
--- a/model/web_rca/v1/incidents_resource.model
+++ b/model/web_rca/v1/incidents_resource.model
@@ -29,7 +29,7 @@ resource Incidents {
     in ResponsibleManagerId String
     in ParticipantId String
     in Mine Boolean
-    // in Status String - TODO: This causes problems, figure out why
+    in Status String
     in IncidentName String
     out Total Integer
     out Items []Incident


jhernand:
You aren't doing anything wrong. `status` happen to be a built-in name used by the generated code, that is why the field is renamed to `status_`, to avoid the conflict. The `Status` function should not have been renamed. This is a bug in the metamodel I am fixing it.

jhernand:
@djberg96 the issue should be addressed in https://github.com/openshift-online/ocm-sdk-go/pull/616.

djberg96:
@jhernand still seeing it: https://gist.github.com/djberg96/3ed0556f8bf84f7368bec7f7f51982cf

jhernand:
OK, am trying to fix that, thanks @djberg96 .

jhernand:
The problem that you see now is the same, but in the server side code. I do not wish to maintain that code because it was experimental and never finished or used. So I am removing it completely.

djberg96:
Excellent, thanks!

PR #540, opened 2022-04-10
Add Import method to HTPasswd users collection

LeshemNoa:
Adding new endpoint to add multiple users to an HTPasswd IDP.

### DIFF
--- a/model/clusters_mgmt/v1/htpasswd_users_resouce.model
+++ b/model/clusters_mgmt/v1/htpasswd_users_resouce.model
@@ -37,6 +37,24 @@ resource HTPasswdUsers {
 		in out Body HTPasswdUser
 	}
 
+	// Adds multiple new users to the _HTPasswd_ file.
+	method Import {
+	  // List of users to add to the IDP.
+      in Items []HTPasswdUser
+
+      // Index of the requested page, where one corresponds to the first page.
+      in out Page Integer = 1
+
+      // Number of items contained in the returned page.
+      in out Size Integer = 100
+
+      // Total number of items of the collection.
+      out Total Integer
+
+      // Updated list of users of the IDP.
+      out Items []HTPasswdUser
+    }
+
 	// Reference to the service that manages a specific _HTPasswd_ user.
 	locator HtpasswdUser {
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/htpasswd_users_resouce.model
```diff
@@ -37,6 +37,24 @@ resource HTPasswdUsers {
 		in out Body HTPasswdUser
 	}
 
+	// Adds multiple new users to the _HTPasswd_ file.
+	method Import {
+	  // List of users to add to the IDP.
+      in Items []HTPasswdUser
+
+      // Index of the requested page, where one corresponds to the first page.
+      in out Page Integer = 1
```
jhernand, line 46: What are the meaning of "Page" and "Size" as inputs here? I'd say that they aren't needed. I'd say they aren't needed as output either. And same for "Total". The only thing that is needed here is "Items", and it only needs to be input:

```
method Import {
        // List of users to add to the IDP:
       in Items []HTPasswdUsers
}
```

If the user wants to get the updated list of users of the IDP she can use get "Get" method.

LeshemNoa: I thought it made sense to return the updated collection, since we're breaking restful API rules anyway. I think it's useful. 

Besides, I think that it's a little bit more consistent with the existing API: the newly created user is returned when they post to the collection, and what we're actually doing here is recreating the collection, so it made sense to me to return that collection.

If that still doesn't sit with you then I'll update the API.

jhernand: No object, but you need to put the "Items" parameter in one line for both input and output:

```
in out Items []HTPasswdUser
```

LeshemNoa: I think it's helpful to make this distinction because it allows me to leave different comments. This can help SDK consumers understand that the `in Items` refers to the new users to be imported, and that the `out Items` refers to the updated collection. It also shows up nicely in the IDE.

PR #541, opened 2022-04-11
Fix JSON representation of log severity

jhernand:


jhernand:
@Lir10 apparently the server doesn't accept the all lower case values that the SDK generates from these names. Is that correct? If so then this patch can be used to force the SDK to use the right names. Please review.

Lir10:
@jhernand OSL validates the severity types here https://gitlab.cee.redhat.com/service/ocm-service-log/-/blob/master/pkg/api/service_log_types.go#L55 

We don't use the SDK for the validation

jhernand:
OK, I see there that the names that the server expects are indeed `Debug`, `Info`, etc, starting with upper case. This patch is necessary to simplify use of that from other applications. I am merging it. Thanks  @Lir10. For the future remember that you can use the `@json(name = "...")` annotation to solve this kind of issue.

PR #542, opened 2022-04-11
Add missing STS fields

zgalor:


zgalor:
@pvasant PTAL
I saw this comment in CS API so wasn't sure if we should add this field ATM:
```
	// TODO: Deprecate CustomIAMRoles in favour of InstanceIAMRoles
	CustomIAMRoles     *CustomIAMRoles `json:"custom_iam_roles,omitempty"`
```

zgalor:
@tzvatot PTAL

tzvatot:
LGTM

PR #558, opened 2022-04-28
add support for api listening in managed service

renan-campos:


model/service_mgmt/v1/cluster_type.model
```diff
@@ -31,4 +31,5 @@ struct Cluster {
 	AWS AWS
 	Nodes ClusterNodes
 	Properties [String]String
+	Api ClusterAPI
```
ciaranRoche, line 34: Should this be all caps :thinking:  Api -> API

ciaranRoche:
lgtm

PR #567, opened 2022-05-18
Add `hypershift.enabled` to cluster type

jhernand:
This patch adds the `hypershift.enabled` flag to the cluster type. This
will be used to indicate if the cluster should be created using
`HyperShift`.

Related: Related: https://issues.redhat.com/browse/SDE-1586


jhernand:
FYI: @samira-barouti

PR #585, opened 2022-07-11
Added registry credential delete method

tirthct:
**Changed**
- Added delete method in registry_credential_resourse.model

**Tested**
- ocm-sdk-go should build successfully : tested

### DIFF
--- a/model/accounts_mgmt/v1/registry_credential_resource.model
+++ b/model/accounts_mgmt/v1/registry_credential_resource.model
@@ -20,4 +20,8 @@ resource RegistryCredential {
 	method Get {
 		out Body RegistryCredential
 	}
+
+	// Delete the registry credentials
+	method Delete {
+	}
 }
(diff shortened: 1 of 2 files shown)


model/accounts_mgmt/v1/registry_credential_resource.model
```diff
@@ -20,4 +20,8 @@ resource RegistryCredential {
 	method Get {
 		out Body RegistryCredential
 	}
+
+	// Delete the registry credentials
```
cristianoveiga, line 24: Since this deletes single registry credential, I would change the comment to "Delete the registry credential"

cristianoveiga:
@tirthct looks good to me. 

Please update the CHANGES.md file as well.

model/accounts_mgmt/v1/registry_credential_resource.model
tirthct: Changed

cristianoveiga:
LGTM

PR #601, opened 2022-08-04
Add hypershift config to the provision shard API and change the management cluster from name to ID

tzvatot:
@samira-barouti @zgalor @pvasant PTAL

### DIFF
--- a/model/clusters_mgmt/v1/provision_shard_type.model
+++ b/model/clusters_mgmt/v1/provision_shard_type.model
@@ -38,9 +38,12 @@ class ProvisionShard {
 	// Contains the cloud provider name
 	link CloudProvider CloudProvider
 
-	// Contains the name of the management cluster for Hypershift clusters that are assigned to this shard.
+	// Contains the configuration for Hypershift
+	HypershiftConfig ServerConfig
+
+	// Contains the OCM cluster ID of the management cluster for Hypershift clusters that are assigned to this shard.
 	// This field is populated by OCM, and must not be overwritten via API.
-	ManagementCluster String
+	ManagementClusterID String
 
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/provision_shard_type.model
```diff
@@ -38,9 +38,12 @@ class ProvisionShard {
 	// Contains the cloud provider name
 	link CloudProvider CloudProvider
 
-	// Contains the name of the management cluster for Hypershift clusters that are assigned to this shard.
+	// Contains the configuration for Hypershift
+	HypershiftConfig ServerConfig
```
zgalor, line 42: comment should end with `.` 

Also the one above (cloud provider name), which is not part of this PR...

tzvatot: non of the comments in this file ends with a dot.

tzvatot: Fixed. PreTAL

zgalor:
LGTM. Why is this WIP?

model/clusters_mgmt/v1/provision_shard_type.model
samira-barouti: Is this the ServerConfig of the Service Cluster? HypershiftConfig is a bit confusing as it suggest that this is the config of the hosted hypershift cluster. Something along the line ServiceClusterConfig might be better.

tzvatot: Given that CS is cluster service, naming something "service cluster" is very confusing IMO.

From OCM perspective, there are two types of provisioning shards - hive, and hypershift.

That's also the name of the field in CS model and DB.

PR #611, opened 2022-08-23
Update ClusterRegistration type

tbrisker:


tbrisker:
@pvasant PTAL, you added these fields a couple of years ago :)

model/clusters_mgmt/v1/cluster_registration_type.model
```diff
@@ -15,30 +15,38 @@ limitations under the License.
...
-	//
-	// ```json
-	// {
-	//   "external_id": "d656aecf-11a6-4782-ad86-8f72638449ba"
-	// }
-	// ```
 	ExternalID String
+
+	// Optional Display Name for the cluster.
+	DisplayName String
```
ciaranRoche, line 48: Just curious, what is the status with DisplayName, I was under the impression we had removed DisplayName from CS or has that change been reverted?

tbrisker: it was reverted

ciaranRoche: Ah cool cool cool

pvasant: @tbrisker @ciaranRoche  we dont want to use display name it was reverted in CS until AMS fixes their logic. however we want new users to stop using the display name . can you please remove it

PR #615, opened 2022-09-06
Trivial: fix space on model

lucasponce:
Hi,
I'm starting to review the repo.
Could I send minor fixes like this one?

Some projects prefer not to add them unless there is a bigger feature/change included; just let me know.

Thanks!

### DIFF
--- a/model/clusters_mgmt/v1/cluster_type.model
+++ b/model/clusters_mgmt/v1/cluster_type.model
@@ -199,8 +199,8 @@ class Cluster {
 	// It is enabled by default
 	DisableUserWorkloadMonitoring Boolean
 
-  // Contains information about Managed Service
-  ManagedService ManagedService
+    // Contains information about Managed Service
+    ManagedService ManagedService
 
 	// HyperShift configuration.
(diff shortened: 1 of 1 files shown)


lucasponce:
I've noticed that the `make lint` on this project fails.
So perhaps the indentation is not enforced?

tbrisker:
@lucasponce i'm not sure GH is displaying the spacing correctly here, it seems to still be off? might be a matter of tabs vs. spaces.

renan-campos:
Tomer's right: the other fields are tabbed whereas this field is using spaces. It may be that in your IDE they appear aligned. Sorry to say this mistake originates from me :/ since then I changed my vim settings so that I can now tell different whitespace apart.

renan-campos:
Oh, I was going to create a code suggestion that replaces the space with the tab, but I don't think its possible in github...

lucasponce:
Perhaps this one is outdated.
I'll try to fix indentation issues on regular PR.
This one as proposed when I was getting familiar with the code.

PR #621, opened 2022-09-21
Added expiry setting to managed service cluster

renan-campos:


renan-campos:
@ciaranRoche  Actually, I just realized we might not need to merge this: MS already knows how to read the expired_at field from the json body. WDYT?

ciaranRoche:
> @ciaranRoche Actually, I just realized we might not need to merge this: MS already knows how to read the expired_at field from the json body. WDYT?

There is no harm in adding it, for the SDK

PR #633, opened 2022-10-31
vpc.aws_subnets: Add missing `public` field

cben:
@pvasant please review

### DIFF
--- a/model/clusters_mgmt/v1/aws_subnetwork_type.model
+++ b/model/clusters_mgmt/v1/aws_subnetwork_type.model
@@ -22,6 +22,9 @@ struct Subnetwork {
 	// Name of the subnet according to its `Name` tag on AWS
 	Name String
 
+	// Whether it is a public subnet
+	Public Boolean
+    
 	// The availability zone to which the subnet is related
 	AvailabilityZone String
 }


ciaranRoche:
LGTM

@cben do you have merge rights on this or do you need me to merge it?

cben:
Nope, I don't have merge rights.

Say, what's the process until changes here become visible in the OpenAPI on api.openshift.com?  
- [ ] an ocm-go-sdk release?
- [ ] upgrading clusters-service to use newer SDK release?
- [ ] a deploy of clusters-service?
- Or is OpenAPI built indepently of the SDK? And is it deployed independently?

Is there an easy way for us to generate latest openapi from master of this repo, without pressuring you to hurry & deploy them (as I expect a stream of similar tiny changes...)

ciaranRoche:
As far as I am aware, openapi is built off this. So changes to master will be reflected in stage, and new releases will be reflected in prod api.openshift.com. 

The SDK is different, so to get the changes in there, you need to bump the model version and generate the sdk from that, using this doc https://github.com/openshift-online/ocm-sdk-go/blob/main/CONTRIBUTING.md

CS only needs to be updated with the SDK for its integration tests.

cben:
Cool, sounds easier that I thought :relieved:  
Although I see no change yet in https://api.stage.openshift.com/, `Subnetwork` has no `public` field :question: 

I don't see anything that would generate & deploy openapi under .github/workflows/.  
This repo's README merely says:
> This documentation is used to automatically generate OpenAPI reference documentation.

if you find out more, would be nice to add details on how/where it gets deployed...

Anyway, I found `make openapi` here, which that immediately produced JSON with new `public` field :heavy_check_mark:, so that what I'll document for UI team to consume definitions merged to `main`.

PR #640, opened 2022-11-09
Add Hypershift field for CloudRegion resource to the model

andreadecorte:
Related: SDA-7081

### DIFF
--- a/model/clusters_mgmt/v1/cloud_region_type.model
+++ b/model/clusters_mgmt/v1/cloud_region_type.model
@@ -37,4 +37,7 @@ class CloudRegion {
 	// 'true' if the region is supported only for CCS clusters, 'false' otherwise.
 	CCSOnly Boolean
 
+	// 'true' if the region is supported for Hypershift deployments, 'false' otherwise.
+	SupportsHypershift Boolean
+
 }


tzvatot:
LGTM

@zgalor @oriAdler @vkareh @ciaranRoche  I wonder about the field name. Initially "HypershiftEnabled" sounded solid as we have "Enabled" field which is more general. But then I noticed we have "SupprtsMultiAZ" so I'm wondering flavour we should take moving forward. Any recommendations?

zgalor:
> LGTM
> 
> @zgalor @oriAdler @vkareh @ciaranRoche I wonder about the field name. Initially "HypershiftEnabled" sounded solid as we have "Enabled" field which is more general. But then I noticed we have "SupprtsMultiAZ" so I'm wondering flavour we should take moving forward. Any recommendations?

Enabled, and Supports usualy mean different things.
The fact that a feature is supported doesn't mean it's enabled.
If I understand the meaning of that field I think `SupportsHypershift` is more appropriate here

andreadecorte:
> > LGTM
> > @zgalor @oriAdler @vkareh @ciaranRoche I wonder about the field name. Initially "HypershiftEnabled" sounded solid as we have "Enabled" field which is more general. But then I noticed we have "SupprtsMultiAZ" so I'm wondering flavour we should take moving forward. Any recommendations?
> 
> Enabled, and Supports usualy mean different things. The fact that a feature is supported doesn't mean it's enabled. If I understand the meaning of that field I think `SupportsHypershift` is more appropriate here

I agree, that was my initial choice too as it is not something to be enabled in this context. I'll be bold and switch this back to SupportsHypershift in the PR

tzvatot:
> > > LGTM
> > > @zgalor @oriAdler @vkareh @ciaranRoche I wonder about the field name. Initially "HypershiftEnabled" sounded solid as we have "Enabled" field which is more general. But then I noticed we have "SupprtsMultiAZ" so I'm wondering flavour we should take moving forward. Any recommendations?
> > 
> > 
> > Enabled, and Supports usualy mean different things. The fact that a feature is supported doesn't mean it's enabled. If I understand the meaning of that field I think `SupportsHypershift` is more appropriate here
> 
> I agree, that was my initial choice too as it is not something to be enabled in this context. I'll be bold and switch this back to SupportsHypershift in the PR

Looks like the "Supports" is the wording also used in AWS API. For example: https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html

PR #644, opened 2022-11-21
Add Capabilities locator in root resource model

tirthct:


model/accounts_mgmt/v1/root_resource.model
```diff
@@ -134,4 +134,8 @@ resource Root {
         target CloudResources
     }
 
+	// Reference to the resource that manages the collection of capabilities.
```
cristianoveiga, line 137: is this indentation a little off?

cristianoveiga:
LGTM

model/accounts_mgmt/v1/root_resource.model
tirthct: Fixed

PR #657, opened 2022-12-21
[SDE-1882] Add bounded service account signing key to STS resource

gdbranco:
Related issue: https://issues.redhat.com/browse/SDE-1882


gdbranco:
@tzvatot PTAL

PR #661, opened 2022-12-28
Add bounded service account key kms id to Sts model

gdbranco:
Related issue: https://issues.redhat.com/browse/SDE-1882

PR #662, opened 2022-12-28
Add bounded service account key kms id to Sts model

gdbranco:
Related issue: https://issues.redhat.com/browse/SDE-1882


gdbranco:
@zgalor PTAL

Ended up closing #661 and opening this one, I had done --amend but for a different reason and didn't notice I didn't had my own commit at the top

PR #670, opened 2023-01-12
SDA-7306 Update NodePool model with status attributes

lucasponce:


model/clusters_mgmt/v1/node_pool_type.model
```diff
@@ -38,4 +41,7 @@ class NodePool {
 
     // The subnet upon which the nodes are created.
     Subnet String
+
+    // Adds additional information about the NodePool status when Replicas != CurrentReplicas.
+    Message String
```
tzvatot, line 46: Perhaps this should be a struct?
Also it sounds very generic. Can you provide an example of what this msg will be?

lucasponce: The message is open.
NodePools don't have a lifecycle; the status is the CurrentReplicas; this message is a custom message that would explain the situation.

It's generic. Then a string is the best option for this case.

tbrisker (commented):
I don't see this in the backend yet, normally we update the api model after the backend is merged

model/clusters_mgmt/v1/node_pool_type.model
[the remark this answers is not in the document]
lucasponce: I don't think so.
There is no lifecycle.
So, the strategy I'm following in this iteration is to explain why the CurrentReplicas don't match the desired one.

If a former "status" is defined in hypershift then we can reevaluate, but I'd prefer not to introduce an artificial lifecycle.

lucasponce:
> I don't see this in the backend yet, normally we update the api model after the backend is merged

Ah good point, thanks for the pointer. 
I'll hold on this PR until the backend is good.

model/clusters_mgmt/v1/node_pool_type.model
lucasponce: But +1.
It makes sense if status grows.
Thanks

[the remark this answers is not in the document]
lucasponce: I was biased in my first thought as I was mimicking other structures.
But yes, let's wrap the message inside a status struct.

lucasponce:
cc @vkareh 

This are the changes needed in the model to update the ROSA CLI with the new NodePool status.

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.0.250 Jan 12 2023
+- Add `NodePoolStatus` type to the `NodePool` resource model.
```
vkareh, line 6: We usually add these during the release tagging, after the actual changes have been merged, so you should probably remove this change.

model/clusters_mgmt/v1/node_pool_status_type.model
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright (c) 2022 Red Hat, Inc.
```
vkareh, line 2: 2023

model/clusters_mgmt/v1/node_pool_type.model
```diff
@@ -23,6 +23,9 @@ class NodePool {
     // Replicas and autoscaling cannot be used together.
     Replicas Integer
 
+    // The number of Machines (and Nodes) currently created.
+    CurrentReplicas Integer
```
vkareh, line 27: This is outdated, remove.

vkareh:
LGTM

lucasponce:
Not yet ready, it requires an update in the Cluster Status.

ciaranRoche:
lgtm

PR #679, opened 2023-01-30
Add Private Link Configuration related resources

andreadecorte:
This is based on this design doc: https://docs.google.com/document/d/1o_aZsW4L8UO2Qu0-k-m24-Y9K_DA3axzQzd79IX01Ps/edit?usp=sharing 

I defined a different type for cluster payload as day2 structure is not compatible (no link), so we can't share

Related: SDA-7802


model/clusters_mgmt/v1/private_link_cluster_configuration_type.model
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the configuration for the Private Links.
+struct PrivateLinkClusterConfiguration {
+	  // List of additional principals for the Private Link
```
zgalor, line 19: looks like an indentation issue

model/clusters_mgmt/v1/private_link_configuration_principal_resource.model
```diff
@@ -0,0 +1,28 @@
...
+resource PrivateLinkPrincipal {
+	// Retrieves the details of the principal.
+	method Get {
+		out Body PrivateLinkPrincipal
+	}
+
+	// Deletes the principal.
+	method Delete {
+	}
+}
```
zgalor, line 27: No `Add` method?

model/clusters_mgmt/v1/cluster_resource.model
```diff
@@ -155,4 +155,8 @@ resource Cluster {
 	locator Hypershift {
 		target Hypershift
 	}
+
+	locator PrivateLinkConfiguration {
```
zgalor, line 159: If it's part of the aws struct, shouldn't it be under `.../clusters/id/aws/private_link_configuration`?

model/clusters_mgmt/v1/private_link_configuration_principals_type.model
```diff
@@ -0,0 +1,21 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Contains a list of principals for the Private Link.
+class PrivateLinkPrincipals {
```
zgalor, line 18: I think this type is obsolete. The resource is enough

model/clusters_mgmt/v1/private_link_cluster_configuration_type.model
```diff
@@ -0,0 +1,23 @@
...
+    Principals []PrivateLinkPrincipal
```
zgalor, line 20: Why aren't you using `PrivateLinkPrincipals`?

zgalor:
What's the difference between: `PrivateLinkClusterConfiguration` and `PrivateLinkConfiguration` ?

andreadecorte:
> What's the difference between: `PrivateLinkClusterConfiguration` and `PrivateLinkConfiguration` ?

PrivateLinkConfiguration contains a link
> link Principals PrivateLinkPrincipals
So it doesn't contain the Principals itself. So it can't be reused for cluster.aws and I defined a specific type for this use case called PrivateLinkClusterConfiguration

I don't like it very much, but for our design I don't see a way to reuse the same struct in the 2 cases

model/clusters_mgmt/v1/cluster_resource.model
andreadecorte: but this one is to generate endpoint is /<clusterid>/private_link_configuration, so locator is under cluster.

model/clusters_mgmt/v1/private_link_cluster_configuration_type.model
andreadecorte: because that would mean that to use it we need to do 
`privateLinkConfiguration.Principals().Principals().Items()`
with this we can remove a useless level and do just 
`privateLinkConfiguration.Principals().Items()`
which is nicer

zgalor:
@tzvatot mind taking a look?
I am not very comfortable with this, but I don't have better suggestions

model/clusters_mgmt/v1/cluster_resource.model
zgalor: So you need another layer of `aws_resource` under cluster_resource, and place the locator there

zgalor:
@andreadecorte I now understand the design better. In general LGTM, but just add another layer of AWS resource between the cluster and the private link configuration

model/clusters_mgmt/v1/cluster_resource.model
andreadecorte: makes sense, done and updated design doc

zgalor:
LGTM

PR #680, opened 2023-01-30
Update changes to 0.0.253

rawsyntax:
For changes in https://github.com/openshift-online/ocm-api-model/pull/678


tylercreller:
LGTM

PR #681, opened 2023-01-30
[SDA-7890] Add oidc and host byo oidc config resources

gdbranco:
New endpoints under /hosted_oidc_configs
Includes:
* List
* Add
* /id Get
* /id Delete

### DIFF
--- a/model/clusters_mgmt/v1/hosted_oidc_config_resource.model
+++ b/model/clusters_mgmt/v1/hosted_oidc_config_resource.model
@@ -1,0 +1,27 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a hosted oidc configuration.
+resource HostedOidcConfig {
+	// Retrieves the details of a HostedOidcConfig
+	method Get {
+		out Body HostedOidcConfig
+	}
+
+	// Deletes the HostedOidcConfig.
+	method Delete {
+	}
+}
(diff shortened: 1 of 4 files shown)


model/clusters_mgmt/v1/host_byo_oidc_config_resource.model
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages hosting of byo oidc configuration.
+resource HostConfig {
```
zgalor, line 18: Is there a single host config, or do we allow multiple configs.

The way it is defined now you will have 

```
POST /oidc/host_config --body {...}

GET /oidc/host_config 

```

So you post and get to the same endpoint (which we don't usually do). What happens if you post and there is already a host config? error/override?

We usually have a collection, and post a new member to it

gdbranco: Why do I have the get in this? I expected Add to be a post only

model/clusters_mgmt/v1/oidc_resource.model
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages hosting of byo oidc configuration.
+resource Oidc {
```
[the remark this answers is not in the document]
gdbranco: Didn't notice this. Will fix it

model/clusters_mgmt/v1/host_byo_oidc_config_resource.model
zgalor: So users can't view current config?

gdbranco: They can download it from the oidc endpoint url that will be a response. Viewing the current config was not a pre requisite and they can't do so at the moment either.

gdbranco: > Is there a single host config, or do we allow multiple configs.

The host config is created based on the private key secret arn. In theory a client only needs one, but it is able to host more than one

gdbranco: Please check https://github.com/openshift-online/ocm-sdk-go/pull/721/files/3e6896abac7ec571b89de8636d9b0cc014a89050#diff-976f5f720f13a78e02eb86ce309c52f04d9575f62b12967ecbbc40571dc2c97a

@zgalor

ciaranRoche (approved):
lgtm

PR #706, opened 2023-03-17
[SDA-7890] Adjust resources to match /oidc_configs endpoints

gdbranco:


model/clusters_mgmt/v1/oidc_configs_resource.model
hunterkepley, line 1: For these comments, should they still include `hosted`?

hunterkepley: for the top 2 comments that is

model/clusters_mgmt/v1/oidc_config_resource.model
```diff
@@ -15,13 +15,18 @@ limitations under the License.
 */
 
 // Manages a hosted oidc configuration.
```
hunterkepley, line 17: Is this still always a `hosted` oidc configuration?

gdbranco: Oh missed it somewhere

PR #718, opened 2023-04-04
Openapi model fix

ciaranRoche:
This change contains a number of fixes for the openapi model. The openapi model created for clusters_mgmt currently is not a valid openapi model, due to missing parameter elements and circular dependencies.

First is bumping the metamodel version, this is to include a vital change to the openapi which is generator, to include the field `required: true` for path parameters.

The second change is removing  the circular dependency from clusters mgmt, as we have a number of objects which reference the cluster object. Despite not using these references at the api level.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,9 @@
 #
 
 # Details of the metamodel used to check the model:
-metamodel_version:=v0.0.54
+metamodel_version:=v0.0.57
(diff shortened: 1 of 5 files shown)


gdbranco:
LGTM

lucasponce (approved):

PR #729, opened 2023-04-23
Add inflight checks locator to cluster resource

zgalor:
Related: SDA-8933


zgalor:
@oriAdler @tbrisker @tzvatot PTAL

PR #732, opened 2023-05-04
feat: Makefile support OSX and Windows

leseb:
Prior to this patch, the models could only be built on Linux. Now
Windows and OSX are supported.
The Make build will fail if the OS is not Windows, Linux or OSX.

Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/.github/workflows/check-pull-request.yaml
+++ b/.github/workflows/check-pull-request.yaml
@@ -22,13 +22,19 @@ on:
     - main
 
 jobs:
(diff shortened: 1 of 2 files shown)


leseb:
@gdbranco @ciaranRoche PTAL 🙂

ciaranRoche (approved):
/lgtm 

Verified and tested locally, works on fedora :smile_cat:

PR #735, opened 2023-05-08
added AWS to Cloud_region_type.model

hunterkepley:
SDA-8783 is  related


model/clusters_mgmt/v1/cloud_region_type.model
```diff
@@ -40,4 +40,6 @@ class CloudRegion {
 	// 'true' if the region is supported for Hypershift deployments, 'false' otherwise.
 	SupportsHypershift Boolean
 
+	// AWS credentials for checking if cloud region exists when POSTing
```
ciaranRoche, line 43: Why do we need to pass credentials? I dont think this will scale, as we will need to pass creds for all regions here not just AWS

PR #738, opened 2023-05-17
Changing parameter name from HttpTokensState to Ec2MetadataHttpTokens

tsorya:
Changing parameter name from HttpTokensState to Ec2MetadataHttpTokens


renan-campos (approved):
\lgtm This aligns with the conversation on the ticket.

PR #739, opened 2023-05-17
Revert "Changing parameter name from HttpTokensState to Ec2MetadataHttpTokens"

renan-campos:
Reverts openshift-online/ocm-api-model#738

Because a release of ROSA has been made with the API using the original parameter name, we'll need to coordinate a new release before making this change so that the feature is not broken.


renan-campos:
I'm reverting the change for now to not block other changes from being released. Sorry for the commit log noise.

PR #744, opened 2023-05-18
Fixing ec2-metadata-http-tokens field

tsorya:


renan-campos:
checked the api and sdk

PR #750, opened 2023-05-24
OCM-15 : Add MachineTypeRootVolume to MachinePool

leseb:
This will allow us to set a Size and/or IOPS for a given root volume.

### DIFF
--- a/model/clusters_mgmt/v1/machine_pool_type.model
+++ b/model/clusters_mgmt/v1/machine_pool_type.model
@@ -44,4 +44,7 @@ class MachinePool {
 
 	// List of security groups to be applied to MachinePool (Optional)
 	SecurityGroupFilters []MachinePoolSecurityGroupFilter
+
+	// The machine root volume capabilities.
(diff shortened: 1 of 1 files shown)


leseb:
@ciaranRoche @tzvatot @gdbranco PTAL

leseb:
Another option if we don't want to use the AWS struct is to put `RootVolume` in `MachinePool` like so:

```
class MachinePool {
...
	// The machine root volume capabilities.
	RootVolume MachineTypeRootVolume
...
...
}
```


Thoughts?

leseb:
I remember @vkareh saying that `AWSMachinePool` and `AWS` structs were generally used for specific parameters. Perhaps this could tilt the balance in favor of `MachineTypeRootVolume`.

tzvatot:
LGTM

@vkareh LGTY?

vkareh:
LGTM

romfreiman:
hmm why it does not merge?

PR #754, opened 2023-05-25
Add Load balancer type to Ingress model

srampal:
@vkareh @ciaranRoche @gdbranco PTAL and approve if ok


model/clusters_mgmt/v1/load_balancer_flavor_type.model
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+// Type of node received via telemetry.
+enum LoadBalancerFlavor {
+	// Classic Load Balancer.
+	Classic
+
+	// Network Load Balancer.
+	Nlb
```
gdbranco, line 23: Should this be NLB or is it ok to keep it as Nlb?

srampal: I thought about making it NLB but wanted to check of the model processing functionality will then convert it to something like "n_l_b" rather than "nlb" in the JSON which is what we want. I can do either .. just need to double check

gdbranco: Oh I see, you can generate a local sdk with these changes by `make generate model_version="HEAD" model_url="<path_to_local_api_model_folder>"`

model/clusters_mgmt/v1/load_balancer_flavor_type.model
```diff
@@ -0,0 +1,24 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Type of node received via telemetry.
+enum LoadBalancerFlavor {
+	// Classic Load Balancer.
+	Classic
```
ciaranRoche, line 20: can we add the following annotations to these to ensure the value is as expected 

```
@json(name = "classic")
```

srampal: Done .. PTAL and approve/ merge if okay.

ciaranRoche (approved):

PR #767, opened 2023-06-12
OCM-16: add root volume capabilities to cluster nodes

leseb:
Add ComputeRootVolume into ClusterNodes to reflect the root volume capabilities of a cluster node.


leseb:
@tzvatot @ciaranRoche @zgalor PTAL. Thanks!

zgalor (approved):

PR #769, opened 2023-06-14
OCM-844|feat : Enable users to provide both hashed and plain-text passwords

ckandag:
Ref: https://issues.redhat.com/browse/OCM-844

currently CS only accepts plain-text passwords.

HTPassword files have passwords already hashed,  For rosacli or ui to support accepting user data from HTpassword files as input, It needs to be able to indicate to CS that the passwords has already been hashed so it doesnt try to rehash them.

This also allows for other scenarios where user does not wish to send passwords in plain-text to the rest API


ciaranRoche (approved):
/lgtm

PR #775, opened 2023-06-22
Document more about CloudRegion

cben:
These fields has existed on cloud_providers/..., aws_inquiries/regions, gcp_inquiries/regions APIs for a long time, but missing in openapi & SDK.

- @vkareh I don't know what else we can explain about `govcloud`?  IIUC from https://github.com/openshift/rosa/pull/702, govcloud regions will always have `"enabled": false` in main api.openshift.com, to actually use them one would need an entirely different API deployment?

  Anyway for me in UI, the `"enabled": false, "govcloud": true` combination is useful (UI will start showing some disabled regions with a reason why disabled; checking this field allows the 2 govcloud ones to remain hidden)

- Added `kms_location_id`, `kms_location_name` (the latter tricky to use imho, added suggestion to mail ocm-feedback if someone wants to rely on it).

- Covered `cloud_providers?fetchRegions=true` mode.

### DIFF
--- a/model/clusters_mgmt/v1/cloud_region_type.model
+++ b/model/clusters_mgmt/v1/cloud_region_type.model
@@ -43,4 +43,13 @@ class CloudRegion {
 	// Whether the region is an AWS GovCloud region.
 	@json(name = "govcloud")
 	GovCloud Boolean
+
+	// (GCP only) Comma-separated list of KMS location IDs that can be used with this region.
+	// E.g. "global,nam4,us". Order is not guaranteed.
+	KMSLocationID String
+
+	// (GCP only) Comma-separated list of display names corresponding to KMSLocationID.
+	// E.g. "Global,nam4 (Iowa, South Carolina, and Oklahoma),US". Order is not guaranteed but will match KMSLocationID.
+	// Unfortunately, this API doesn't allow robust splitting - Contact ocm-feedback@redhat.com if you want to rely on this.
+	KMSLocationName String
 }
--- a/model/clusters_mgmt/v1/cloud_region_type.model
+++ b/model/clusters_mgmt/v1/cloud_region_type.model
@@ -28,7 +28,7 @@ class CloudRegion {
 	// Link to the cloud provider that the region belongs to.
 	link CloudProvider CloudProvider
 
-	// Whether the region is enabled for deploying an OSD cluster.
+	// Whether the region is enabled for deploying a managed cluster.
 	Enabled Boolean
 
 	// Whether the region supports multiple availability zones.
@@ -40,4 +40,7 @@ class CloudRegion {
 	// 'true' if the region is supported for Hypershift deployments, 'false' otherwise.
 	SupportsHypershift Boolean
 
+	// Whether the region is an AWS GovCloud region.
+	@json(name = "govcloud")
+	GovCloud Boolean
 }
--- a/model/clusters_mgmt/v1/cloud_provider_type.model
+++ b/model/clusters_mgmt/v1/cloud_provider_type.model
@@ -22,4 +22,7 @@ class CloudProvider {
 	// Name of the cloud provider for display purposes. It can contain any characters,
 	// including spaces.
 	DisplayName String
+
+	// (optional) Provider's regions - only included when listing providers with `fetchRegions=true`.
+	Regions []CloudRegion
 }
--- a/model/clusters_mgmt/v1/cloud_providers_resource.model
+++ b/model/clusters_mgmt/v1/cloud_providers_resource.model
@@ -54,6 +54,10 @@ resource CloudProviders {
 		// results is undefined.
 		in Order String
 
+		// If true, includes the regions on each provider in the output. Could slow request response time.
+		@http(name = "fetchRegions")
+		in FetchRegions Boolean
+
 		// Total number of items of the collection that match the search criteria,
 		// regardless of the size of the page.
 		out Total Integer


cben:
- [ ] Draft until #776 and https://github.com/openshift-online/ocm-api-metamodel/pull/197 settle
- [ ] will need to update `fetchRegion` accordingly.

cben:
Rebased on top of #779 since `@http(name = "fetchRegions")` added here also requires metamodel 0.0.59.

cben:
- [ ] wait, it's `govcloud` not `gov_cloud`

zgalor:
@cben What is the motivation for using the `@http` annotation?

cben:
The query param backend accepts here is `?fetchLabels=true`, with lowercase first letter.  Similar to several other such params (#776), but not quite following API guidelines.

Just naming it `in fetchLabels Boolean` did not work right, metamodel really wants field names  to be CamelCase (in part so that they match public Go names).  
But we can override the name sent over the wire (and declared in OpenAPI) with `@http` and/or `@json`;  we'd previously used `@json` in similar cases but after conversation with @jhernand he concluded `@http` is more appropriate for query params and did some fixes https://github.com/openshift-online/ocm-api-metamodel/pull/197 to support that (see diffs I posted there for exact effects it had on OpenAPI and SDK code).

cben:
- [ ] umm let me rebase again, this PR should contain fetchRegions but not fetchLabels etc, those already landed in #779.

cben:
PTAL, this is only about regions now.

model/clusters_mgmt/v1/cloud_region_type.model
```diff
@@ -40,4 +40,16 @@ class CloudRegion {
 	// 'true' if the region is supported for Hypershift deployments, 'false' otherwise.
 	SupportsHypershift Boolean
 
+	// Whether the region is an AWS GovCloud region.
+	@json(name = "govcloud")
+	GovCloud Boolean
+
+	// (GCP only) Comma-separated list of KMS location IDs that can be used with this region.
+	// E.g. "global,nam4,us". Order is not guaranteed.
+	KMSLocationID String
```
zgalor, line 49: Can the metamodel parse this or should it be `KmsLocationId`?

cben: Good question, lemme check... :heavy_check_mark: Yes it parses it fine:

openapi:
```json
          "kms_location_id": {
            "description": "(GCP only) Comma-separated list of KMS location IDs that can be used with this region.\nE.g. \"global,nam4,us\". Order is not guaranteed.",
            "type": "string"
          },
          "kms_location_name": {
            "description": "(GCP only) Comma-separated list of display names corresponding to KMSLocationID.\nE.g. \"Global,nam4 (Iowa, South Carolina, and Oklahoma),US\". Order is not guaranteed but will match KMSLocationID.\nUnfortunately, this API doesn't allow robust splitting - Contact ocm-feedback@redhat.com if you want to rely on this.",
            "type": "string"
          },
```
go:
```go
		stream.WriteObjectField("kms_location_id")
		stream.WriteString(object.kmsLocationID)
             ...
		stream.WriteObjectField("kms_location_name")
		stream.WriteString(object.kmsLocationName)
             ...

// KMSLocationID returns the value of the 'KMS_location_ID' attribute, or
// the zero value of the type if the attribute doesn't have a value.
//
// (GCP only) Comma-separated list of KMS location IDs that can be used with this region.
// E.g. "global,nam4,us". Order is not guaranteed.
func (o *CloudRegion) KMSLocationID() string {
  ...

// GetKMSLocationID returns the value of the 'KMS_location_ID' attribute and
// a flag indicating if the attribute has a value.
//
// (GCP only) Comma-separated list of KMS location IDs that can be used with this region.
// E.g. "global,nam4,us". Order is not guaranteed.
func (o *CloudRegion) GetKMSLocationID() (value string, ok bool) {
  ...
  
// KMSLocationName returns the value of the 'KMS_location_name' attribute, or
// the zero value of the type if the attribute doesn't have a value.
//
// (GCP only) Comma-separated list of display names corresponding to KMSLocationID.
// E.g. "Global,nam4 (Iowa, South Carolina, and Oklahoma),US". Order is not guaranteed but will match KMSLocationID.
// Unfortunately, this API doesn't allow robust splitting - Contact ocm-feedback@redhat.com if you want to rely on this.
func (o *CloudRegion) KMSLocationName() string {
  ...

// GetKMSLocationName returns the value of the 'KMS_location_name' attribute and
// a flag indicating if the attribute has a value.
//
// (GCP only) Comma-separated list of display names corresponding to KMSLocationID.
// E.g. "Global,nam4 (Iowa, South Carolina, and Oklahoma),US". Order is not guaranteed but will match KMSLocationID.
// Unfortunately, this API doesn't allow robust splitting - Contact ocm-feedback@redhat.com if you want to rely on this.
func (o *CloudRegion) GetKMSLocationName() (value string, ok bool) {
  ...

// KMSLocationID sets the value of the 'KMS_location_ID' attribute to the given value.
//
//
func (b *CloudRegionBuilder) KMSLocationID(value string) *CloudRegionBuilder {
  ...

// KMSLocationName sets the value of the 'KMS_location_name' attribute to the given value.
//
//
func (b *CloudRegionBuilder) KMSLocationName(value string) *CloudRegionBuilder {
  ...
```

zgalor (approved):
LGTM

PR #776, opened 2023-06-22
Use `@http` instead of `@json` for `dryRun` parameter

jhernand:
Currently we are using the `@json` annotation to force the name of the dry run query parameter to `dryRun`. But this is incorrect, and only works because of a bug in the metamodel that is going to be fixed. This patch replaces `@json` with `@http` which will work after the bug is fixed.

Related: https://github.com/openshift-online/ocm-api-metamodel/pull/197

### DIFF
--- a/model/addons_mgmt/v1/addon_resource.model
+++ b/model/addons_mgmt/v1/addon_resource.model
@@ -24,7 +24,7 @@ resource Addon {
 	// Updates the addon.
 	method Update {
 		// DryRun indicates the request body will not be persisted when dryRun=true. 
-		@json(name = "dryRun")
+		@http(name = "dryRun")
 		in DryRun Boolean
 
 		in out Body Addon
--- a/model/addons_mgmt/v1/addon_version_resource.model
+++ b/model/addons_mgmt/v1/addon_version_resource.model
@@ -24,7 +24,7 @@ resource AddonVersion {
 	// Updates the addon version.
 	method Update {
 		// DryRun indicates the request body will not be persisted when dryRun=true. 
-		@json(name = "dryRun")
+		@http(name = "dryRun")
 		in DryRun Boolean
 
 		in out Body AddonVersion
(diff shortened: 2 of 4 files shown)


cben:
LGTM.

Well technically POST with `?method=get` does pass same params as keys of JSON body.  
But requiring both `@json` and `@http` would be too brittle; IMHO single annotation should determine naming in both normal GET (where they go via URL) and POST where they they are json.  

And I think `@http` is the more logical one to use, as in all other methods they are URL params, and because conceptually our `?method=get` can be thought of as "alternative transport for http GET semantics" :+1:

cben:
Hmm but wait.  `in out` params go by URL to the server but return by JSON body back, right?

I'm now uncertain which annotation is better.

cben:
after chatting yesterday, LGTM :+1:.

- I agree that for `in out` params it's fine to write both `@json` and `@http` annotations to be explicit.
- AFAICT, metamodel & the SDK are not aware at all of POST `?method=get` hack?  It's accepted in clusters_mgmt server for all GET routes (by api.MethodMiddleware) but not documented in openapi, and I don't see the SDK making (or even allowing?) such POST requests.  
  So not a relevant consideration.

https://github.com/openshift-online/ocm-api-metamodel/pull/197 implemented a fallback such that `@json` (in absense of `@http`) can also affect URL params — but my understanding that's temporary for backward-compatibility, and this change is still desirable.

jhernand:
@cben you are right in all your comments.

Would be nice to have support in the SDK for `?method=get` but I think that nobody will invest time to do it (including myself) so don't count on that.

cben (approved):

cben:
@tirthct @etabak Apparently my approve doesn't count here, so needs another review.

This is similar to #779. While many of these are POST taking a json body, `dryRun` is sent as a query param.
Backend logic was implemented https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2193 :
```go
type CreateArguments struct {
	DryRun bool `json:"dryRun" default:"false"`
}
```

```go
	query := api.CreateArguments{}
	err := UnmarshalRequestArgs(r, &query)
```
(and UnmarshalRequestArgs() [does reflection on `json:` struct tag](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/ce4bd1ebfdb7af4221ad47c0430726bf0b25b24c/cmd/clusters-service/servecmd/apiserver/handler_helpers.go#L62))

PR #777, opened 2023-06-22
OCM-844 | feat: Add HashedPassword field  to provide encrypted value

ckandag:
Instead of overloading Password field and using a  boolean to indicate the encryption status Adding a seperate HashedPassword field to make it more apparent to the user

Ref: https://issues.redhat.com/browse/OCM-844


ciaranRoche (approved):
/lgtm

PR #780, opened 2023-06-23
Ocm 926 : Adding DeleteAssociatedResources parameter to accounts delete endpoint

tirthct:
**Changed**
- Adding DeleteAssociatedResources parameter to accounts delete endpoint

**Tested**
- Parameter should be added to the endpoint : tested

### DIFF
--- a/model/accounts_mgmt/v1/account_resource.model
+++ b/model/accounts_mgmt/v1/account_resource.model
@@ -27,7 +27,7 @@ resource Account {
 	}
 
 	method Delete {
-	    in DeleteAssociatedResources Boolean
+	    in DeleteAssociatedResources Boolean = false
 	}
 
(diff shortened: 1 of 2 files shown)


model/accounts_mgmt/v1/account_resource.model
```diff
@@ -27,6 +27,7 @@ resource Account {
 	}
 
 	method Delete {
+	    in DeleteAssociatedResources Boolean
```
tylercreller, line 30: let's set a default like cluster_resource

tirthct: Set deault to false

tylercreller:
LGTM

tylercreller (approved):

tylercreller (approved):

PR #781, opened 2023-06-25
OCM-209 | feat: add autoscaler API resources

osherdp:
After discussing about it, we decided that instead of handling all cluster-autoscaler parameters via requests to the ``cluster`` resource, we should allow the editing functionality only by the handling of the
``/api/clusters_mgmt/v1/clusters/{cluster_id}/autoscaler`` endpoint.

We're still interested in setting autoscaler parameters in POST requests to the cluster object, as it's a great help to some consumers (most importantly the OCM UI).

This means:
* We need to add a resource file in the meta-model, to make sure the SDK has a way to build requests against it. The URI enables ``POST, GET, DELETE, PUT`` for the singleton resource (``/api/clusters_mgmt/v1/clusters/{cluster_id}/autoscaler``. No special ID it allocated for the autoscaler resource).
* The autoscaler is no longer a struct but rather a class, as it now has an actual identity.
* The ``cluster`` class only contains a link to the ``autoscaler`` resource. This was made to spare doing database joins, as well as making it more clear to API users that setting cluster-autoscaler values is *not* possible when doing PATCH requests to the cluster API resource.

This changes the original plan we had defined at https://github.com/openshift-online/ocm-api-model/pull/747


osherdp:
cc @zgalor @tzvatot @ciaranRoche

ciaranRoche:
@osherdp is this discussion/decision documented anywhere in jira or a doc etc?

osherdp:
> @osherdp is this discussion/decision documented anywhere in jira or a doc etc?

I'm intending to add it to the design document after we have our meeting today with product and UI

ciaranRoche:
> > @osherdp is this discussion/decision documented anywhere in jira or a doc etc?
> 
> I'm intending to add it to the design document after we have our meeting today with product and UI

Can you add me to that meeting please :pray:

osherdp:
DDR also changed to reflect the intended changes

tzvatot (approved):
LGTM

model/clusters_mgmt/v1/autoscaler_request_type.model
```diff
@@ -0,0 +1,41 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Contains the necessary attributes to allow creation of cluster autoscalers
+struct ClusterAutoscalerRequest {
```
tzvatot, line 18: Why `request`? we usually just write the name of the type.

tzvatot (changes requested):
Added a comment

model/clusters_mgmt/v1/autoscaler_request_type.model
osherdp: Everywhere I'm seeing we're doing this in the POST-requests definition:
```
method Post {
    in Request <object>Request
```

there are not a lot of cases where we write something else (I could only find ``SubscriptionRegistration`` and ``LabelRequestPayload``)

tzvatot: I can't comment about why AMS is using "X-Request" pattern. But for CS, we don't use this convention. The only exception that I found is STSCredentialsRequest, but that's because the endpoint itself contains "request" in the URL, so it's aligns with our conventions.

osherdp: ack. I changed it to ``ClusterAutoscaler``

tzvatot (approved):

tzvatot:
LGTM

PR #822, opened 2023-08-28
Implement new account reference type 

dagbay-rh:
## Summary

* Create a new `AccountReference` type and link it to `Subscription.Creator`
* Remove `Name` attribute on `Account`

## Why

`Subscription.Creator` returns an `AccountReference` per the AMS api docs: https://api.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/get_api_accounts_mgmt_v1_subscriptions

Currently an `Account` is being returned instead, which results in some info being lost when the AMS response is decoded into a Subscription object. One disrepancy is that `Account` contains `first_name` and `last_name`, while `AccountReference` contains `name`. So that means right now the name of the creator is lost when calling accounts mgmt get subscriptions.

I left in the existing `Creator` field of type `Account` for backwards compatibility purposes. I can remove that if requested or update things however needed for best practices

Slack thread for full context on this change: https://redhat-internal.slack.com/archives/CBDNMS43V/p1692729609296909

## Example

when querying for subscriptions at `/api/accounts_mgmt/v1/subscriptions/<sub id>?fetchAccounts=true`, ams returns a creator field that is a reference to the account that created the subscription

this field looks like this in the raw json response:
```
...
  "creator": {
    "email": "dagbay@redhat.com",
    "href": "/api/accounts_mgmt/v1/accounts/2P4XpdoEpHVoPgwfJ9imEN5HOfE",
    "id": "2P4XpdoEpHVoPgwfJ9imEN5HOfE",
    "kind": "Account",
    "name": "Daniel Agbay",
    "username": "rh-ee-dagbay"
  },
...
```

right now this gets unmarshalled into an `Account` object, which does not contain the `name` field, so that data gets lost

### DIFF
--- a/model/accounts_mgmt/v1/accout_reference_type.model
+++ b/model/accounts_mgmt/v1/accout_reference_type.model
@@ -15,7 +15,6 @@ limitations under the License.
 */
 
 class AccountReference {
-    Kind	    string
     Email	    string
     Name	    string
     Username	string
--- a/model/accounts_mgmt/v1/accout_reference_type.model
+++ b/model/accounts_mgmt/v1/accout_reference_type.model
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2019 Red Hat, Inc.
+Copyright (c) 2023 Red Hat, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dagbay-rh:
this is not working currently. when building ocm-sdk with this change I get an error:
```
➜  ocm-sdk-go git:(main) ✗ make
cd examples && \
for i in *.go; do \
	go build ${i} || exit 1; \
done
go: downloading golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f
go: downloading golang.org/x/text v0.3.8
# github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1
../accountsmgmt/v1/subscription_type_json.go:467:8: duplicate case "creator" (constant of type string) in expression switch
	../accountsmgmt/v1/subscription_type_json.go:463:8: previous case
make: *** [Makefile:34: examples] Error 1
```

so duplicating the `creator` field in the json representation of `Subscription.Creator` breaks the subscription marshaler. Not sure how to solve this since we need to read `creator` into `Account` and `AccountReference` for backwards compatibility, but as is this breaks the marshalers

dagbay-rh:
we are going to make changes in AMS to return `first_name` and `last_name` instead of `name` on the `AccountReference` type. So this means no changes in the sdk will be required.

PR #823, opened 2023-08-29
OCM-1759 | feat: add CIDR block to /vpcs output

thomasmckay:


gdbranco:
Don't know why it doesn't have it but could you include a final '.' for each sentence?

thomasmckay:
> Don't know why it doesn't have it but could you include a final '.' for each comment?

Do you mean treat the comments as full sentences? If so, I can update the entries in these two files.

gdbranco (approved):

PR #825, opened 2023-08-29
revert pr 817

dagbay-rh:
revert adding name to account type from https://github.com/openshift-online/ocm-api-model/pull/817

further investigations/discussions have lead us to question the best approach here so we are reverting these changes for now. slack thread with more context: https://redhat-internal.slack.com/archives/CBDNMS43V/p1692729609296909


cristianoveiga (approved):

cristianoveiga:
LGTM

PR #827, opened 2023-09-01
feat: add Topology field on ServerConfig model

lucasponce:


andreadecorte:
we should also add 
ProvisionShardDedicatedTopology = "dedicated"

model/clusters_mgmt/v1/server_config_type.model
```diff
@@ -21,4 +21,7 @@ class ServerConfig {
 
 	// The kubeconfig of the server
 	Kubeconfig String
+
+	// The topology of a provision shard (Optional)
```
gdbranco, line 25: Not sure why nothing on this file has the '.', but we usually required it. Do you know if there's some reason for this file not to have that for the comments?

gdbranco (approved):

model/clusters_mgmt/v1/server_config_type.model
lucasponce: Ah, true, I've created the file from scratch, and I forgot.

I'll update it

tbrisker (approved):

PR #828, opened 2023-09-01
OCM-3559 - Feat : Add DisplayName and Description to billing_model_item_type.model

tirthct:
**Changed**
- Add DisplayName and Description to billing_model_item_type.model

**Tested**
- DisplayName and Description properties should be visible through ocm sdk : tested


cristianoveiga:
LGTM

PR #831, opened 2023-09-06
OCM-3599: added image_overrides to version type

cristianoveiga:


model/clusters_mgmt/v1/billing_model_item_type.model
```diff
@@ -0,0 +1,36 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// BillingModelItem represents a billing model
+struct BillingModelItem {
```
tzvatot, line 18: I wonder if this should be class? class have ID and HREF and Kind by default.

tzvatot (changes requested):

model/clusters_mgmt/v1/billing_model_item_type.model
cristianoveiga: Yes, it makes sense ... I copied it over from the accounts_mgmt folder, but I believe we better use the class in this case. 

I will updated it in both sides. cc: @tirthct

cristianoveiga: @tzvatot I've replaced the struct it with a class. PTAL

renan-campos (approved):
/lgtm

model/accounts_mgmt/v1/billing_model_item_type.model
```diff
@@ -1,15 +1,30 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// BillingModelItem represents a billing model
+class BillingModelItem {
+	// BillingModelType is the type of the BillingModel. e.g. standard, marketplace
```
zgalor, line 19: missing period at the end of each remark

zgalor (approved):
/lgtm

PR #848, opened 2023-10-12
Revert "Revert "OCM-1534: add backplane_url field in the environment struct.""

machi1990:
Reverts openshift-online/ocm-api-model#840

We can now go ahead with the change. 

Reverting this in preparation for the implementation

/cc @tzvatot


ciaranRoche (approved):
/lgtm

PR #878, opened 2023-12-08
OCM-5156 | add search and order param to nodepools

chenz4027:
https://issues.redhat.com/browse/OCM-5156

### DIFF
--- a/model/clusters_mgmt/v1/node_pools_resource.model
+++ b/model/clusters_mgmt/v1/node_pools_resource.model
@@ -16,61 +16,60 @@ limitations under the License.
 
 // Manages the collection of node pools of a cluster.
 resource NodePools {
-    // Retrieves the list of node pools.
-    method List {
-        // Index of the requested page, where one corresponds to the first page.
-        in out Page Integer = 1
+	// Retrieves the list of node pools.
+	method List {
+		// Index of the requested page, where one corresponds to the first page.
+		in out Page Integer = 1
 
-        // Number of items contained in the returned page.
-        in out Size Integer = 100
+		// Number of items contained in the returned page.
+		in out Size Integer = 100
 
-        // Search criteria.
-		// 
+		// Search criteria.
+		//
 		// The syntax of this parameter is similar to the syntax of the _where_ clause of a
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/node_pools_resource.model
```diff
@@ -24,6 +24,37 @@ resource NodePools {
         // Number of items contained in the returned page.
         in out Size Integer = 100
 
+        // Search criteria.
+		// 
+		// The syntax of this parameter is similar to the syntax of the _where_ clause of a
+		// SQL statement, but using the names of the attributes of the cluster instead of
```
andreadecorte, line 30: it would be good to have a real example valid for node pools, like we do here https://github.com/openshift-online/ocm-api-model/blob/main/model/clusters_mgmt/v1/pending_delete_clusters_resource.model#L29

model/clusters_mgmt/v1/node_pools_resource.model
```diff
@@ -24,6 +24,37 @@ resource NodePools {
...
+		// name like 'my%' and region.id = 'us-east-1'
+		// ```
+		//
+		// If the parameter isn't provided, or if the value is empty, then all the
+		// clusters that the user has permission to see will be returned.
+		in Search String
+
+		// Order criteria.
+		//
+		// The syntax of this parameter is similar to the syntax of the _order by_ clause of
```
andreadecorte, line 45: same here

model/clusters_mgmt/v1/node_pools_resource.model
```diff
@@ -24,6 +24,37 @@ resource NodePools {
         // Number of items contained in the returned page.
         in out Size Integer = 100
 
+        // Search criteria.
+		//
```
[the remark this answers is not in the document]
chenz4027: Yep the file was using space instead of the tab. Fixed.

model/clusters_mgmt/v1/node_pools_resource.model
chenz4027: Added the example of id desc.

andreadecorte (approved):

PR #884, opened 2024-01-09
Modify notify_details response

Lir10:


Lir10:
@etabak @zgalor ptal

tzvatot:
@Lir10 should this be reflected in https://api.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/post_api_accounts_mgmt_v1_notify_details? I don't see these fields currently

Lir10:
@etabak yes, we are working on merging it in AMS

tzvatot:
> @tzvatot it was merged can you please merge this MR ?

@Lir10  the API seems to not be consistent with the general API, where subscription is a struct, and under that you have the fields of "id", "kind" and "href".

Lir10:
@tzvatot sure, this is not a standard API .. it's used specifically for OSL to get some data. there's no need for the other fields

model/accounts_mgmt/v1/generic_notify_details_response_type.model
```diff
@@ -0,0 +1,41 @@
...
+// class that defines notify details response in general.
+class GenericNotifyDetailsResponse {
+	// Index of the returned page, where one corresponds to the first page. As this
+	// collection doesn't support paging the result will always be `1`.
+	Page Integer
+
+	// Number of items that will be contained in the returned page. As this collection
+	// doesn't support paging or searching the result will always be the total number of
+	// notify details.
+	Size Integer
```
tzvatot, line 26: If this is not a searchable collection, why do we need page, size, and total? why can't we just have the Items field alone?

tzvatot (dismissed):

Lir10:
created a ticket https://issues.redhat.com/browse/OCM-5561

model/accounts_mgmt/v1/generic_notify_details_response_type.model
tzvatot: Will be handled in https://issues.redhat.com/browse/OCM-5561

PR #892, opened 2024-01-28
OCM-5791 | feat: Add a locater for `/aws_inquiries/vpcs/{vpc_id}`

oriAdler:


oriAdler:
We won't be adding a new endpoint with a URL fragment.

PR #895, opened 2024-01-29
OCM-5802 Add external_auth_config to cluster

lucasponce:
It adds the ExternalAuthConfig struct under the cluster model.


chenz4027:
@lucasponce I think we might need to add the change log in the CHANGES.md

ciaranRoche (approved):

PR #909, opened 2024-03-04
OCM-6454 | feat: add component routes attribute

gdbranco:
Related issue: OCM-6454

### DIFF
--- a/model/clusters_mgmt/v1/component_route_kind_type.model
+++ b/model/clusters_mgmt/v1/component_route_kind_type.model
@@ -1,0 +1,22 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Type of Component Route.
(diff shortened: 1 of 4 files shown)


oriAdler (approved):

model/clusters_mgmt/v1/component_route_kind_type.model
```diff
@@ -0,0 +1,30 @@
...
+enum ComponentRouteType {
+	// Oauth.
+	@json(name = "oauth")
+	Oauth
+
+	// Downloads.
+	@json(name = "downloads")
+	Downloads
+
+    // Console.
```
tzvatot, line 27: Looks like indentation is off here

tzvatot (changes requested):

model/clusters_mgmt/v1/component_route_kind_type.model
```diff
@@ -0,0 +1,30 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Type of Component Route.
+enum ComponentRouteType {
+	// Oauth.
```
tzvatot, line 19: This documentation seems redundant as it does not provide any addition value other than the field name. I suggest either add some doc or remove the line completely.

model/clusters_mgmt/v1/component_route_kind_type.model
```diff
@@ -0,0 +1,30 @@
...
+	@json(name = "oauth")
```
tzvatot, line 20: Do we need the json field mapping? I think it would be the default as the field name in lowercase.

tzvatot (changes requested):

ciaranRoche (approved):
/lgtm

PR #910, opened 2024-03-04
Add node drain grace period to the NodePool model

philipwu08:
Link to Jira: https://issues.redhat.com/browse/OCM-6101

DDR: https://docs.google.com/document/d/1aqCcPL9JIW9LjCh9ojfu5yJnpjv8mXRoPsS2lGWON0g

### DIFF
--- a/model/clusters_mgmt/v1/node_pool_type.model
+++ b/model/clusters_mgmt/v1/node_pool_type.model
@@ -50,4 +50,7 @@ class NodePool {
 
     // The names of the tuning configs for this node pool.
     TuningConfigs []String
+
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/node_pool_type.model
```diff
@@ -50,4 +50,7 @@ class NodePool {
 
     // The names of the tuning configs for this node pool.
     TuningConfigs []String
+
+    // Time to wait on upgrades for a NodePool to drain before it is removed.
```
andreadecorte, line 54: It is also for replace operations, not only upgrades

gdbranco (approved):

PR #912, opened 2024-03-06
OCM-4459 | fix: change casing to fix the create failure issue

chenz4027:
https://issues.redhat.com/browse/OCM-4459

### DIFF
--- a/model/clusters_mgmt/v1/external_auth_type.model
+++ b/model/clusters_mgmt/v1/external_auth_type.model
@@ -84,6 +84,7 @@ struct ExternalAuthClaim {
 struct TokenClaimMappings {
 
     // Username is a name of the claim that should be used to construct usernames for the cluster identity.
+    @json(name = "username")
     UserName UsernameClaim
 
 	// Groups is a name of the claim that should be used to construct groups for the cluster identity.


model/clusters_mgmt/v1/external_auth_type.model
```diff
@@ -84,7 +84,7 @@ struct ExternalAuthClaim {
 struct TokenClaimMappings {
 
     // Username is a name of the claim that should be used to construct usernames for the cluster identity.
-    UserName UsernameClaim
+    Username UsernameClaim
```
machi1990, line 87: You could also use the json annotation
```
 @json(name = "username")
 UserName UsernameClaim
```
This will ensure backward compatible for the `UserName` field, assuming that the sdk as already been release.

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.0.362 Mar 06 2024
+- Changed `UserName` attribute for TokenClaimMappings to `Username`.
+
```
machi1990, line 7: These changes are typically done on a separate MR once the main PR to introduce the changes has been merged; https://github.com/openshift-online/ocm-sdk-go/blob/main/CONTRIBUTING.md#releasing-a-new-ocm-api-model-version

model/clusters_mgmt/v1/external_auth_type.model
andreadecorte: I am not aware of anyone using this yet so it should be safe. But we can also change the json annotation if we want

machi1990: Yes. And if there were people using it, it's probably not working so we should be fine just changing it now. I wanted to share the above as mostly an FYI

CHANGES.md
chenz4027: Thanks @machi1990 , will remove this commit

model/clusters_mgmt/v1/external_auth_type.model
chenz4027: Thanks @machi1990 @andreadecorte , let me add the json annotation

chenz4027:
@andreadecorte @machi1990 Would you mind taking another look? Thanks!

machi1990 (approved):
I'll let @andreadecorte give this another look and final approval as the SME

chenz4027:
Hmm seems like it's not auto merged. @andreadecorte @machi1990 Do you have merge permission by any chance?

machi1990:
@chenz4027 it is merged now, you can followup with the rest of the changes

PR #925, opened 2024-04-09
[WIP] OCM-7144 | Add model for storage quota values resource

cgiradkar:
The CS backend supported /storage_quota_values endpoint but the OCM sdk did not. Hence ocm-api-model needed to be updated.


model/clusters_mgmt/v1/storage_quota_type.model
```diff
@@ -0,0 +1,24 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of a storage quota 
+class StorageQuota {
```
ahitacat, line 18:
```suggestion
struct StorageQuota {
```

In this case, we can use a struct instead of a class. Class objects will always have an id and href which the StorageQuotaValues doesn't have.

ahitacat (changes requested):

model/clusters_mgmt/v1/storage_quota_type.model
cgiradkar: Changed through commit amend

mnecas (approved):

PR #926, opened 2024-04-10
OCM-7027 | feat: document pagination and ordering support for break glass

andreadecorte:
Break glass credentials now support pagination and ordering, so we document it properly


lucasponce:
LGTM

lucasponce (approved):

PR #934, opened 2024-04-22
OCM-6715 | add user defined tags to machinepool model

willkutler:
https://issues.redhat.com/browse/OCM-6715

### DIFF
--- a/model/clusters_mgmt/v1/aws_machine_pool_type.model
+++ b/model/clusters_mgmt/v1/aws_machine_pool_type.model
@@ -27,4 +27,14 @@ class AWSMachinePool {
 
     // Associates nodepool availability zones with zone types (e.g. wavelength, local).
     AvailabilityZoneTypes [String]String
+
+    // Optional keys and values that the machine pool provisioner will add as AWS tags to all AWS resources it creates.
+    //
+    // AWS tags must conform to the following standards:
+    // - Each resource may have a maximum of 25 tags
+    // - Tags beginning with "aws:" are reserved for system use and may not be set
+    // - Tag keys may be between 1 and 128 characters in length
+    // - Tag values may be between 0 and 256 characters in length
+    // - Tags may only contain letters, numbers, spaces, and the following characters: [_ . : / = + - @]
+    Tags [String]String
 }
--- a/model/clusters_mgmt/v1/aws_node_pool_type.model
+++ b/model/clusters_mgmt/v1/aws_node_pool_type.model
@@ -23,6 +23,13 @@ class AWSNodePool {
     InstanceProfile String
 
     // Optional keys and values that the installer will add as tags to all AWS resources it creates.
+    //
+    // AWS tags must conform to the following standards:
+    // - Each resource may have a maximum of 25 tags
+    // - Tags beginning with "aws:" are reserved for system use and may not be set
+    // - Tag keys may be between 1 and 128 characters in length
+    // - Tag values may be between 0 and 256 characters in length
+    // - Tags may only contain letters, numbers, spaces, and the following characters: [_ . : / = + - @]
     Tags [String]String
 
     // Associates nodepool subnets with AWS Outposts.


model/clusters_mgmt/v1/machine_pool_type.model
```diff
@@ -33,6 +33,9 @@ class MachinePool {
 	// The subnets upon which the nodes are created.
 	Subnets []String
 
+	// Optional keys and values that the installer will add as tags to all AWS resources it creates.
+    Tags [String]String
```
gdbranco, line 37: Spacing seems a bit off

model/clusters_mgmt/v1/machine_pool_type.model
```diff
@@ -33,6 +33,9 @@ class MachinePool {
 	// The subnets upon which the nodes are created.
 	Subnets []String
 
+	// Optional keys and values that the installer will add as tags to all AWS resources it creates.
```
gdbranco, line 36:
```suggestion
	// Optional keys and values that the machine pool provisioner will add as tags to all AWS resources it creates.
```

model/clusters_mgmt/v1/machine_pool_type.model
```diff
@@ -33,6 +33,9 @@ class MachinePool {
 	// The subnets upon which the nodes are created.
 	Subnets []String
 
+	// Optional keys and values that the machine pool provisioner will add as tags to all AWS resources it creates.
+	Tags [String]String
```
miguelsorianod, line 37: Another question: Is the design of a `MachinePool` in CS to be intended to be agnostic of cloud provider? If so, adding here at this level would be a bit problematic because we are going to put AWS MachinePool specific attributes for the generic MachinePool concept

gdbranco: This is right, the tags are included within the aws machine pool spec internally
This should be within [this](https://github.com/openshift-online/ocm-api-model/blob/ac7f60aa7d99e43ba2ca33d33a0af5c36ea4fc23/model/clusters_mgmt/v1/aws_machine_pool_type.model#L18) file indeed

model/clusters_mgmt/v1/machine_pool_type.model
```diff
@@ -33,6 +33,9 @@ class MachinePool {
 	// The subnets upon which the nodes are created.
 	Subnets []String
 
+	// Optional keys and values that the machine pool provisioner will add as tags to all AWS resources it creates.
```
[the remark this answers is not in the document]
gdbranco: I think this rarely have the restrictions on the api-model directly, is there some example to understand how the open api spec will be generated?

model/clusters_mgmt/v1/machine_pool_type.model
miguelsorianod: :+1:

model/clusters_mgmt/v1/machine_pool_type.model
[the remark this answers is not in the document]
miguelsorianod: If we cannot put the restrictions in the api-model / openapi directly, which I think we cannot due to the limitations of the metamodel, we should at least document it for the consumers of the API which it was why I was proposing documenting it.

Aside from that, on CS side we will probably want to add at least some of those validations so it fails earlier in the lifecycle

model/clusters_mgmt/v1/machine_pool_type.model
willkutler: thanks! just changed this

model/clusters_mgmt/v1/machine_pool_type.model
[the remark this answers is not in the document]
willkutler: @miguelsorianod I've changed the description to "AWS tags" for more clarity. On the CS side, we already have added validations for these tags: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/7710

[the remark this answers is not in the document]
miguelsorianod: Perfect thanks.

Even with the validations implemented I think there is value in documenting the restrictions so the consumers of the API are aware of them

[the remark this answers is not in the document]
willkutler: that sounds good to me. where would be the best place to document the restrictions? do you mean in this repo or somewhere else?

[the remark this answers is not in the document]
miguelsorianod: In this repository yes.

That's because then the OpenAPI document is generated from the contents defined here.

Documenting it in the API types of the CS repository itself is also a good idea.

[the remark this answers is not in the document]
willkutler: @miguelsorianod done! also added documentation for validations on the node pool tags

model/clusters_mgmt/v1/aws_machine_pool_type.model
```diff
@@ -27,4 +27,14 @@ class AWSMachinePool {
 
     // Associates nodepool availability zones with zone types (e.g. wavelength, local).
     AvailabilityZoneTypes [String]String
+
+    // Optional keys and values that the machine pool provisioner will add as AWS tags to all AWS resources it creates.
+    //
+    // AWS tags must conform to the following standards:
+    // - Each resource may have a maximum of 50 tags
```
gdbranco, line 34:
```suggestion
    // - Each resource may have a maximum of 25 tags
```

There's a soft limit right? AWS itself is 50 though

willkutler: oh right, was going off the AWS guidelines and forgot we had a smaller limit. I'll change that rn

PR #936, opened 2024-04-30
OCM-7306 | feat: Added support for KubeletConfigs to HCP MachinePools

robpblake:
This PR adds in API Model definitions for the `KubeletConfig` endpoints to support custom PIDs limits on HCP clusters. In this PR we add the following:

- Expose the `/api/clusters_mgmt/v1/clusters/{id}/kubelet_configs` endpoint that allows users to create one or more KubeletConfigs for their HCP clusters
   - For Classic Clusters, validation on this endpoint will permit the creation of a single KubeletConfig
- Expose the `/api/clusters_mgmt/v1/clusters/{id}/kubelet_configs/{kubelet_config_id}` endpoint that allows the user to retrieve details of the specific KubeletConfig
- Add the `kubelet_configs` field to the NodePool resource

### DIFF
--- a/model/clusters_mgmt/v1/cluster_resource.model
+++ b/model/clusters_mgmt/v1/cluster_resource.model
@@ -188,6 +188,10 @@ resource Cluster {
 		target TuningConfigs
 	}
 
+	locator KubeletConfigs {
+	    target KubeletConfigs
+	}
+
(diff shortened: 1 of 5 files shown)


ciaranRoche (approved):
/lgtm

PR #943, opened 2024-05-29
OCM-7374 | add AdditionalAllowedPrincipals to aws model

mikeshng:
Add AdditionalAllowedPrincipals to the aws model.
The AdditionalAllowedPrincipals specifies a list of additional allowed principal ARNs to be added to the hosted control plane's VPC Endpoint Service.

Ref: https://issues.redhat.com/browse/OCM-7374

### DIFF
--- a/model/clusters_mgmt/v1/aws_type.model
+++ b/model/clusters_mgmt/v1/aws_type.model
@@ -70,4 +70,7 @@ struct AWS {
 
 	// Additional AWS Security Groups to be added to default control plane machine pool.
 	AdditionalControlPlaneSecurityGroupIds []String
+
+	// Additional allowed principal ARNs to be added to the hosted control plane's VPC Endpoint Service.
+	AdditionalAllowedPrincipals []String
 }


mikeshng:
`CHANGES.md` release related changes will be in a separate PR

model/clusters_mgmt/v1/aws_type.model
```diff
@@ -70,4 +70,7 @@ struct AWS {
 
 	// Additional AWS Security Groups to be added to default control plane machine pool.
 	AdditionalControlPlaneSecurityGroupIds []String
+
+	// Additional allowed principal ARNs to be added to the hosted control plane's VPC Endpoint Service.
+	AdditionalAllowedPrincipals []String
```
philipwu08, line 75: lgtm, but just for thought, for consistency, would "AdditionalControlPlaneAllowedPrincipals" to be long?

mikeshng: https://hypershift-docs.netlify.app/reference/api/#hypershift.openshift.io/v1beta1.AWSPlatformSpec

I think it's better to stay consistent with the Hypershift field name which is the current `AdditionalAllowedPrincipals`.

lucasponce: @philipwu08 is right

And it's something to learn, as we should capture this in the review of the DDR.

But now propagating this change will impact, and also when ROSA Classic supports this feature for parity the field is added.

lucasponce (approved):

PR #944, opened 2024-05-31
feat: add initial ARO-HCP Azure Node Pool attributes

miguelsorianod:
Related to https://issues.redhat.com/browse/ARO-7434

### DIFF
--- a/model/clusters_mgmt/v1/azure_node_pool_type.model
+++ b/model/clusters_mgmt/v1/azure_node_pool_type.model
@@ -1,0 +1,58 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


model/clusters_mgmt/v1/azure_node_pool_type.model
```diff
@@ -0,0 +1,58 @@
...
+    // Required during creation.
+    // Immutable.
+    ResourceName String
+
+    // The Azure Virtual Machine size identifier used for the
+    // Nodes of the Node Pool.
+    // Availability of VM sizes are dependent on the Azure Location
+    // of the parent Cluster.
+    // Required during creation.
+    VMSize String
```
machi1990, line 37: I see that the API model definition in CS has this comment 
```
// VmSize is the Azure Virtual Machine size identifier used for the
	// Nodes of the Node Pool.
	// TODO pending to define if this field is required or not. Availability
	// of VM sizes are dependent on the Azure Location.
	VmSize *string `json:"vm_size,omitempty"`
```
Might be good to update the other part as well if we've confirmed that the field is required

machi1990 (approved):
lgtm

left one comment regarding updating a TODO comment on the CS side api level which is stale if confirmation has been done on the vm_size being required

model/clusters_mgmt/v1/azure_node_pool_type.model
miguelsorianod: It is not yet decided

PR #950, opened 2024-06-26
OCM-7800 | Define and add pagination model to cloud_provider_inquiries

cgiradkar:


model/clusters_mgmt/v1/pagination_type.model
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Pagination contains the data which enables pagination on the response
+struct Pagination {
```
mnecas, line 18: Not that big a fan of this as it is specific only for one model. 
How do other models have the pagination implemented?

cgiradkar: BreakGlassCredentialsListRequest has implemented pagination the request struct like this: https://github.com/openshift-online/ocm-sdk-go/blob/d6662dfa3e339336e2459a413289c5b867500131/clustersmgmt/v1/break_glass_credentials_client.go#L350
I think this way it's more pluggable to any struct that needs to implement pagination on the response.

ziccardi:
lgtm

vkareh:
This seems wrong. We have a `*_resource.model` that automatically adds pagination components using a `method List`. You might want to follow what's being done in https://github.com/openshift-online/ocm-api-model/blob/main/model/clusters_mgmt/v1/break_glass_credentials_resource.model#L20-L62 to support this.

vkareh (changes requested):
This seems wrong. We use a `*_resource.model` that automatically adds pagination components using `method List`. You might want to follow what's being done in https://github.com/openshift-online/ocm-api-model/blob/main/model/clusters_mgmt/v1/break_glass_credentials_resource.model#L20-L62 to support this.

vkareh (commented):
This builds well now and generates the following files in the SDK:

	clustersmgmt/v1/cloud_provider_data_client.go
	clustersmgmt/v1/cloud_provider_data_resource_json.go
	clustersmgmt/v1/cloud_providers_data_client.go
	clustersmgmt/v1/cloud_providers_data_resource_json.go

Is there a specific endpoint in CS associated with this or does it expect the `/{aws,gcp}_inquiries/*` endpoints?

cgiradkar:
> This builds well now and generates the following files in the SDK:
> 
> ```
> clustersmgmt/v1/cloud_provider_data_client.go
> clustersmgmt/v1/cloud_provider_data_resource_json.go
> clustersmgmt/v1/cloud_providers_data_client.go
> clustersmgmt/v1/cloud_providers_data_resource_json.go
> ```
> 
> Is there a specific endpoint in CS associated with this or does it expect the `/{aws,gcp}_inquiries/*` endpoints?

The original task had broader consideration, but this PR only aims to add pagination structure to `gcp_inquiries` endpoints.

PR #971, opened 2024-08-08
ARO-9520: Add state struct to node pool

ahitacat:
This change will add a new struct to store the status of a node pool for v2 API.

We should consider the possibility of joining NodePool.Status.Message and NodePool.Status.State.Detail. 

Added an enum with the possible states of a NodePool, this should be reviewed with future investigations.


tzvatot (approved):
LGTM

PR #974, opened 2024-08-13
ARO-9520: Add UpdatedTimestamp to nodepool state

ahitacat:
This change adds a new field to the state structure to store the timestamp when the state was last changed.


model/clusters_mgmt/v2alpha1/node_pool_state_type.model
```diff
@@ -21,4 +21,8 @@ class NodePoolState {
 
     // Detailed user friendly status for node pool state.
     Details String
+
+    //Timestamp when the the state change las time. Using the format defined in
```
tzvatot, line 25: I believe you need an extra space here before "Timestamp"

PR #981, opened 2024-09-06
OCM-10883 | feat: include new manifest fields

gdbranco:
OCM-10883

### DIFF
--- a/model/clusters_mgmt/v1/manifest_type.model
+++ b/model/clusters_mgmt/v1/manifest_type.model
@@ -18,4 +18,17 @@ limitations under the License.
 class Manifest {
 	// List of k8s objects to deploy on a hosted cluster.
 	Workloads []Interface
+
+	// Spec of Manifest Work object from open cluster management
+	// For more info please check https://open-cluster-management.io/concepts/manifestwork.
+	Spec Interface
+
+	// Date and time when the manifest got created in OCM database.
+	CreationTimestamp Date
+
+	// Date and time when the manifest got updated in OCM database.
+	UpdatedTimestamp Date
+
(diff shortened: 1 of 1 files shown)


model/clusters_mgmt/v1/manifest_type.model
```diff
@@ -18,4 +18,16 @@ limitations under the License.
 class Manifest {
 	// List of k8s objects to deploy on a hosted cluster.
 	Workloads []Interface
+
+	// Spec of Manifest Work object from open cluster management.
```
andreadecorte, line 22: Useful also to add a link to https://open-cluster-management.io/concepts/manifestwork/

model/clusters_mgmt/v1/manifest_type.model
```diff
@@ -18,4 +18,16 @@ limitations under the License.
...
+	Spec Interface
+
+	// Date and time when the manifest got created in clusters-service.
```
andreadecorte, line 25: curious, do we already expose this in the API? Or should we keep something generic like in OCM DB

model/clusters_mgmt/v1/manifest_type.model
gdbranco: Done, thanks!

model/clusters_mgmt/v1/manifest_type.model
gdbranco: What do you mean? We usually expose for other services, not sure why we wouldn't here

ciaranRoche (approved):
/lgtm

andreadecorte (approved):
lgtm

PR #985, opened 2024-09-10
OSD-25124: new WIFOnly flag on version gates

Nikokolas3270:
This flag will be used to filter out version gates applying on OSD (managed) clusters running on GCP provider with WIF (Workload Identity Federation) enabled.


Nikokolas3270:
After discussion:
- Multiplying the `Only` suffixed members is not the way to go. Indeed what if we have to support a similar scheme on azure in the future?
- It was proposed to eventually introduce an `OIDCScheme` member that would take its values in `STS`, `WIF` or the empty string... but doing so looks like labelling the version gate. So actually it may better to extend the use of the existing `Label` member. The label seems to be used for information purposes only for now. What about using it for filtering as well?

PR #1007, opened 2024-12-12
feat: ARO-11963 | add outbound type changes in OCM SDK Go

ziccardi:


model/clusters_mgmt/v1/azure_type.model
```diff
@@ -101,4 +101,8 @@ struct Azure {
 	// Required during creation.
 	// Immutable.
 	OperatorsAuthentication AzureOperatorsAuthentication
+
+	// NodesOutboundConnectivity defines how the network outbound
```
miguelsorianod, line 105: There's some whitespacing issue here

model/clusters_mgmt/v1/azure_nodes_outbound_connectivity_types.model
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// The configuration of the node outbound connectivity
+struct AzureNodesOutboundConnectivity {
+	// OutboundType is the type of network outbound configuration.
+	// The only accepted value is 'load_balancer'
```
miguelsorianod, line 20: We should document that the default is `load_balancer`. Yes, `load_balancer` is the only accepted value (for now), but it should be clear that a default is set for it, the user doesn't explicitly need to set it. I'd also document that it is Immutable

model/clusters_mgmt/v1/azure_type.model
```diff
@@ -101,4 +101,8 @@ struct Azure {
...
+	// configuration of the Cluster's Node Pool's Nodes is performed.
+	NodesOutboundConnectivity AzureNodesOutboundConnectivity
```
miguelsorianod, line 107: Here we should document that by default the network outbound configuration is through an Azure Load Balancer, and that it is immutable.

model/clusters_mgmt/v1/azure_nodes_outbound_connectivity_types.model
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+// The configuration of the node outbound connectivity
+struct AzureNodesOutboundConnectivity {
+	// OutboundType is the type of network outbound configuration.
+	// The default and only accepted value is 'load_balancer'.
+	// This value is immutable.
+	OutboundType String
+}
```
machi1990, line 23:
```suggestion
}

```
nit; missing newline

machi1990 (approved):
left one nit, but lgtm

model/clusters_mgmt/v1/azure_nodes_outbound_connectivity_types.model
ziccardi: fixed!

machi1990 (approved):

PR #1017, opened 2025-01-30
ARO-15561: Add node_pool to arohcp

ahitacat:
Added the resources required for the node_pools endpoint.

```
    "/api/aro_hcp/v1alpha1/clusters/{cluster_id}/node_pools/{node_pool_id}": {
      "delete": {
        "description": "Deletes the node pool.",
        "responses": {
          "202": {
            "description": "Success."
          },
```
OpenAPI and SDK look as expected.

### DIFF
--- a/model/aro_hcp/v1alpha1/node_pool_resource.model
+++ b/model/aro_hcp/v1alpha1/node_pool_resource.model
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2023 Red Hat, Inc.
+Copyright (c) 2025 Red Hat, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,14 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Representation of the status of a node pool.
-class NodePoolStatus {
-    // The current state of the node pool
-    State NodePoolState
-    
-    // The current number of replicas for the node pool.
-    CurrentReplicas Integer
+// Manages a specific nodepool.
+resource NodePool {
+	// Retrieves the details of the node pool.
+	method Get {
+		out Body NodePool
+	}
 
-    // Adds additional information about the NodePool status when the node pool doesn't reach the desired replicas.
-    Message String
+	// Updates the node pool.
+	method AsyncUpdate {
+		in out Body NodePool
+	}
+
+	// Deletes the node pool.
+	method AsyncDelete {
+	}
+
 }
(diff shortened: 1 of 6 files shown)


model/aro_hcp/v1alpha1/node_pool_resource.model
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    // Updates the node pool.
+    method Update {
+        in out Body NodePool
+    }
+
+    // Deletes the node pool.
+    method Delete {
+    }
```
nimrodshn, line 31: Lets keep this with Get only and add methods later.

model/aro_hcp/v1alpha1/node_pools_resource.model
```diff
@@ -0,0 +1,75 @@
...
+
+		// Retrieved list of node pools.
+		out Items []NodePool
+	}
+
+	// Adds a new node pool to the cluster.
+	method Add {
+		// Description of the node pool
+		in out Body NodePool
+	}
```
nimrodshn, line 68: Also here.

nimrodshn:
Can you also add the state fields for the _cmv1_ NodePools so that they get reflected here as well please? :pray:

zgalor:
don't we also need to fix the reference in the cluster resource?

nimrodshn:
> don't we also need to fix the reference in the cluster resource?

No. the compiler knows how to find it _if it exists_ in its service. [e.g. `/aro_hcp/v1alpha1`]

model/aro_hcp/v1alpha1/node_pool_resource.model
ahitacat: I resolved this, but I think it is not needed to remove the rest of the methods. We can hold this PR until we have the enpoints ready.

ahitacat:
This change will require of https://github.com/openshift-online/ocm-api-metamodel/pull/208 to create the SDK.

nimrodshn:
LGTM :100:

model/aro_hcp/v1alpha1/node_pool_resource.model
```diff
@@ -0,0 +1,33 @@
...
+
+// Manages a specific nodepool.
+resource NodePool {
+	// Retrieves the details of the node pool.
+	method Get {
+		out Body NodePool
+	}
+
+	// Updates the node pool.
+	method AsyncUpdate {
```
zgalor, line 25: I lack some context. What's the diff between Update and AsyncUpdate methods?

nimrodshn: The former returns `StatusOK` while the latter returns `StatusAccepted`

nimrodshn: {see previous comment by @ahitacat }

zgalor (approved):

PR #1024, opened 2025-03-03
[WIP] Move generator for clientapi to repo where the model is defined

deads2k:
This also brings in the generator. Taken together this

1. provides generator dependency using standard go.mod versioning
2. allows automated inspection of generator level
3. allows easy injection of testing changes to generator
4. allows testing of generated clientapi at the point we create it
5. keeps existing "generate my entire client locally" flows working
6. Notice that clientapi does NOT depend on its generator

To be done prior to merge

1. go mod vendor verify
2. golang code generator cleaning
3. openapi generator verify
4. golang generator verify
5. proof that ocm-sdk-go works with this generated API
7. proof that a basic unit test can be done in the repo (probably needs another metamodel generator update to specify package with API types).

To be done after merge

1. update generated type to use standard JSON serialization instead of jsoniter.  This will also enable go-fuzz to fuzz our API types
2. write compatiblity testing similar to upstream kube for the client-api


vkareh:
Looks good now, thanks @deads2k 

/hold cancel
/approve
/lgtm

PR #1034, opened 2025-04-24
add verify target and OWNERS in preparation for prow

deads2k:
Also, it appears that `make lint` isn't currently enforced because its failing.  I left it out of this initial pull.


vkareh (approved):
/lgtm

PR #1037, opened 2025-05-15
capture the optional managed identities proposal using API Model.

machi1990:
```
{
 ...rest of cluster spec
 "azure": {
   "required_operators_authentication": {
     "managed_identities": {
       "managed_identities_data_plane_identity_url": "https://dummyhost.identity.azure.net",
       "control_plane_operators_managed_identities": {
         "operator1": {
           "resource_id": "example_resource_id"
         },
       },
       "data_plane_operators_managed_identities": {
         "example_operator": {
           "resource_id": "example_resource_id"
         }
       },
       "service_managed_identity": {
         "resource_id": "example_resource_id"
       }
     }
   },
   "image_registry": {
     "control_plane_identity": {
       "resource_id": "example_resource_id"
       },
     "data_plane_identity": {
       "resource_id": "example_resource_id"
     }
   }
 }
}
```

The PR captures the above proposal as stated out in the DDR around optional managed identities.
However, there are a few challenges with the current proposal;
- if in the future an operator becomes not required all the time anymore, it'll be an API breaking change as we'll have to move  from the `required_operators_authentication` section to its own section as par the current proposal. How do we people see a better design to future proof ourselves? This is the case we're having now with the image registry operator taken as an example. 
- How are generic optional features that requires Managed/Workload identities designed across different clouds? 
The proposal above designs it as `cluster.image_registry` however, the concept of managed identities is tied to Azure. 
Other clouds have different concepts / naming. An alternative to the proposal above will be to have `cluster.azure.image_registry` and `cluster.aws.image_registry` i.e each cloud section in the cluster's spec will have its own type defining of the feature. However, this duplicates the feature definition across different clouds.

Another alternative is  `cluster.image_registry.aws` and `cluster.image_registry.azure` i.e on the feature itself, have configurations per cloud. This duplicates the cloud platform everywhere.

I am opening this discussion so that we can iterate on the design and come up with a solid one to captured  in the DDR and adopted across offerings and specifically for aro-hcp optional features design.

@tzvatot @vkareh @miguelsorianod @deads2k @flavianmissi

### DIFF
--- a/model/clusters_mgmt/v1/azure_type.model
+++ b/model/clusters_mgmt/v1/azure_type.model
@@ -97,10 +97,13 @@ struct Azure {
 	// Immutable.
 	NetworkSecurityGroupResourceID String
 
+    // This is renamed to 'required_operators_authentication'.
+	// It brings concerns of; what if one of the operators stop being required anymore all the time?
+	// How do people see the design considering the future evolutions concerns?
 	// Defines how the operators of the cluster authenticate to Azure.
 	// Required during creation.
 	// Immutable.
-	OperatorsAuthentication AzureOperatorsAuthentication
+	RequiredOperatorsAuthentication AzureOperatorsAuthentication
 
 	// NodesOutboundConnectivity defines how the network outbound
 	// configuration of the Cluster's Node Pool's Nodes is performed.
--- a/model/clusters_mgmt/v1/cluster_image_registry_type.model
+++ b/model/clusters_mgmt/v1/cluster_image_registry_type.model
@@ -1,0 +1,4 @@
+struct ClusterImageRegistry {
+	ControlPlaneIdentity AzureControlPlaneManagedIdentity
+	DataPlaneIdentity AzureDataPlaneManagedIdentity
+}
--- a/model/clusters_mgmt/v1/cluster_type.model
+++ b/model/clusters_mgmt/v1/cluster_type.model
@@ -247,5 +247,11 @@ class Cluster {
 	RegistryConfig ClusterRegistryConfig
 
 	// OpenShift Cluster Capabilities configuration
-	Capabilities ClusterCapabilities
+	// Commented because it'll be removed from the API as par the proposal in the DDR
+	// Capabilities ClusterCapabilities
+
+    // At the moment it is top level at the cluster's spec.
+	// We could move it to the .Azure field as it contains Azure specific things.
+	// This is the one point we want to solve
+	ImageRegistry ClusterImageRegistry
 }


model/clusters_mgmt/v1/cluster_image_registry_type.model
```diff
@@ -0,0 +1,4 @@
+struct ClusterImageRegistry {
+	ControlPlaneIdentity AzureControlPlaneManagedIdentity
```
miguelsorianod, line 2: Another related question we have here is: what happens if the cluster image registry functionality is not supported only on azure but also on aws. We would need to have here something like:
```
ClusterImageRegistry:
AWS:
  ControlPlaneIdentity AWSSTSIdentity
  DataPlaneIdentity AWSSTSIdentity
Azure:
  ControlPlaneIdentity AzureControlPlaneManagedIdentity
  DataPlaneIdentity AzureDataPlaneManagedIdentity
```

or similar

model/clusters_mgmt/v1/azure_type.model
```diff
@@ -97,10 +97,13 @@ struct Azure {
...
 	NetworkSecurityGroupResourceID String
 
+    // This is renamed to 'required_operators_authentication'.
+	// It brings concerns of; what if one of the operators stop being required anymore all the time?
+	// How do people see the design considering the future evolutions concerns?
 	// Defines how the operators of the cluster authenticate to Azure.
 	// Required during creation.
 	// Immutable.
-	OperatorsAuthentication AzureOperatorsAuthentication
+	RequiredOperatorsAuthentication AzureOperatorsAuthentication
```
miguelsorianod, line 106: Even if this remains called OperatorsAuthentication, where you put the core operators information only, the same issue would happen

miguelsorianod:
As a side note:

Even though a PR is a good place where to leave comments, have them tracked and have discussions around them, the fact that we are iterating on the custom ocm api model definitions makes hard to understand how would that *actually* end up looking (as in openapi, go types, ...)

model/clusters_mgmt/v1/azure_type.model
machi1990: +1

model/clusters_mgmt/v1/cluster_image_registry_type.model
vkareh: This is an interesting problem. For completion, the way we've solved this in ROSA is by providing an array of arbitrary operators that require identity:
```
{
  "aws": {
    "operator_iam_roles": [
      {"id": "openshift-image-registry-operator", "role_arn": "<identity resource>"},
    ]
  }
}
```
However, when it comes to any knobs for said operator, they don't live there, they are provided elsewhere on an as-needed basis.

We should probably follow a similar model.

model/clusters_mgmt/v1/azure_type.model
vkareh: The name would matter if you group all identities together, regardless of whether they are for core components or optional components.

vkareh (changes requested):
There are two paths forward here:

1. Bundle all identities; component knobs go in separate structs
- Pros:
  - All identities are in the same location
  - Good separation of concerns: Identities belong to the cloud platform, whereas components belong to the cluster.
  - Single location for validating all identities and providing status
  - More consistent with ROSA
- Cons:
  - Harder to distinguish visually which identities are required

2. Co-locate identity within the discrete component configuration
- Pros:
  - Identities belong where they are used
  - Good separation of configuration
  - Easy to visualize what's required based on components
- Cons:
  - More places for validation/status is cumbersome for client consumption
  - Inconsistent with ROSA

The current option of bundling identities for some components, but co-locating for others is inconsistent and it seems harder to future-proof. Let's not do that and instead let's focus our efforts on improving on one of these two options above. My preference is for the first option, as it's more consistent with how ROSA does it today, but I'm willing to concede based on good arguments for the second option.

model/clusters_mgmt/v1/azure_type.model
miguelsorianod: >  The name would matter if you group all identities together, regardless of whether they are for core components or optional components.

Yes in that case the name wouldn't matter, that's correct. In the master branch that's how the API currently looks like if you want to take a look. This is, all identities are grouped together and there's no differentiation between core and optional components. There is differentiation between control plane and data plane identities because they are conceptually different, configured in a different way and receive/return a different set of data

miguelsorianod:
> There are two paths forward here:
> 
>     1. Bundle all identities; component knobs go in separate structs
> 
> 
>     * Pros:
>       
>       * All identities are in the same location
>       * Good separation of concerns: Identities belong to the cloud platform, whereas components belong to the cluster.
>       * Single location for validating all identities and providing status
>       * More consistent with ROSA
> 
>     * Cons:
>       
>       * Harder to distinguish visually which identities are required
> 
> 
>     2. Co-locate identity within the discrete component configuration
> 
> 
>     * Pros:
>       
>       * Identities belong where they are used
>       * Good separation of configuration
>       * Easy to visualize what's required based on components
> 
>     * Cons:
>       
>       * More places for validation/status is cumbersome for client consumption
>       * Inconsistent with ROSA
> 
> 
> The current option of bundling identities for some components, but co-locating for others is inconsistent and it seems harder to future-proof. Let's not do that and instead let's focus our efforts on improving on one of these two options above. My preference is for the first option, as it's more consistent with how ROSA does it today, but I'm willing to concede based on good arguments for the second option.

I think when considering this it is also relevant to keep in mind cluster upgrades:
When a Cluster is upgraded to an OCP version, aside from changing the `version` attribute in the API payload, if the new OCP version has new required core operators the end user will need to introduce the corresponding managed identity information for them. In the same way, if the new OCP version stops using previously existing core operators the end user would need to remove them when performing the patch. Something similar could happen if for some reason an optional CS functionality stops being supported. Although the latter seems unlikely as it would be a breaking change

machi1990:
Thanks @vkareh @miguelsorianod  for chiming in.

I am for option 2; which is feature/functionality centric. 

**Re: consistency with rosa,** as mentioned in slack thread, rosa-hcp isn't consistency with itself. That is STS iam roles for operators are configured in the STS operators roles however there is a feature (an optional one, a user can chose to activate ir or not) named AuditLog forwarding which follows the design as what's proposed in the DDR; similar to option 2. 

**Re: status of managed identities;** Me, as user I care more of the status of the cluster or the extra functionalities that I've enabled on this cluster. The cluster already has a "status" object which can be enriched to return the "state of the world at time T of managed identities or anything else we would like to report". For example today, as part of  inflight checks on MIs (if the MIs don't exist or anything) the status object would report that. Additionally, the inflight check status endpoint provides additional status of whatever the failure was.  I personally find the `cluster.status` object to be a place where the "state of the world" is reported and not the managed_identities part which is the spec. 

**More places for validation** - this concern is already the case. The cluster payload is big and the user/consumer of the API needs to make sure that they are providing a valid payload.  In regards to validation, some functionality validation are interconnected by the knobs that are associated with that validation. Having knobs related to the functionality in one place helps quiet a lot in this and makes it easier for the user and CS developers themselves to perform those validations. 
e.g the image registry case, if everything is in one spec, it is a matter of validating that image registry spec. However, if one knob on managed identity is in the "managed_identities" spec, the other knob about feature enablement is on "ocp_capabilities" spec, the other xxx knob is on the "image_registry"  related spec, validating this becomes hard in my opinion.

**user wanting to know the managed identities associated to a cluster at a central place while looking at the cluster schema** Me, as a user the API, this is an interesting use case that second option doesn't offer however there are solution to that (e.g a dedicate endpoint similar to what rosa offers that returns the list of all operators roles of a given cluster) if this really becomes an issue from the visualisation point of view. Also this use case, doesn't address the next followup question a user might ask themselves; why do I have this MI? To which feature / functionality is this needed for? The answer can be "for this operator identified by this name " but operator names change or could change - so it is important (from my point of view) that  a user clearly sees the presence of a certain managed identity from functionality/feature standpoint.

miguelsorianod:
> Thanks @vkareh @miguelsorianod for chiming in.
> 
> I am for option 2; which is feature/functionality centric.
> 
> **Re: consistency with rosa,** as mentioned in slack thread, rosa-hcp isn't consistency with itself. That is STS iam roles for operators are configured in the STS operators roles however there is a feature (an optional one, a user can chose to activate ir or not) named AuditLog forwarding which follows the design as what's proposed in the DDR; similar to option 2.
> 
> **Re: status of managed identities;** Me, as user I care more of the status of the cluster or the extra functionalities that I've enabled on this cluster. The cluster already has a "status" object which can be enriched to return the "state of the world at time T of managed identities or anything else we would like to report". For example today, as part of inflight checks on MIs (if the MIs don't exist or anything) the status object would report that. Additionally, the inflight check status endpoint provides additional status of whatever the failure was. I personally find the `cluster.status` object to be a place where the "state of the world" is reported and not the managed_identities part which is the spec.
> 
> **More places for validation** - this concern is already the case. The cluster payload is big and the user/consumer of the API needs to make sure that they are providing a valid payload. In regards to validation, some functionality validation are interconnected by the knobs that are associated with that validation. Having knobs related to the functionality in one place helps quiet a lot in this and makes it easier for the user and CS developers themselves to perform those validations. e.g the image registry case, if everything is in one spec, it is a matter of validating that image registry spec. However, if one knob on managed identity is in the "managed_identities" spec, the other knob about feature enablement is on "ocp_capabilities" spec, the other xxx knob is on the "image_registry" related spec, validating this becomes hard in my opinion.
> 
> **user wanting to know the managed identities associated to a cluster at a central place while looking at the cluster schema** Me, as a user the API, this is an interesting use case that second option doesn't offer however there are solution to that (e.g a dedicate endpoint similar to what rosa offers that returns the list of all operators roles of a given cluster) if this really becomes an issue from the visualisation point of view. Also this use case, doesn't address the next followup question a user might ask themselves; why do I have this MI? To which feature / functionality is this needed for? The answer can be "for this operator identified by this name " but operator names change or could change - so it is important (from my point of view) that a user clearly sees the presence of a certain managed identity from functionality/feature standpoint.

I have a similar thinking to Manyanda and it is what I commented in previous discussions.

For me, if we decide to go to this route though we need to define how would we provide authentication support if the given optional functionality is supported in multiple cloud providers. This is, imagine you have the image registry functionality which is supported in both AWS and Azure based cluster types. In the case of AWS ones an AWS STS identity would be required and in the case of Azure based clusters an Azure Managed Identity would be required. How would this look like at CS API level? this is considering that we currently have an `azure` and an `aws` top level section in the Cluster payload too

vkareh:
@miguelsorianod 
> if the new OCP version stops using previously existing core operators the end user would need to remove them when performing the patch

Not strictly true. The identity would be unnecessary, but not removing it would not affect cluster operation/upgrade. In either case, this seems orthogonal to the shape of the API, the user behaviour would exist in either case.

@machi1990 
> I am for option 2; which is feature/functionality centric.

Yes. I personally like the ergonomics of this approach. However, we need to define it very carefully, since based on @miguelsorianod's comment, it will become awkward for OCP components that require identities and are supported on more than one cloud/platform. That alone is enough to tip me in the other direction.

Can you update this PR to see how it would look in this case?

miguelsorianod:
> Not strictly true. The identity would be unnecessary, but not removing it would not affect cluster operation/upgrade.

:+1: . That's accurate for the case of the removal. However, I think conceptually that's what we should require. I think we shouldn't have identities information in the API payload that are not used nor supported anymore.

>  In either case, this seems orthogonal to the shape of the API, the user behaviour would exist in either case.
The user behavior would exist in either case, but we have to think of how the experience of the user would be, as in what changes in the API payload via a PATCH would need to be performed with a way of organizing the API or the other. Like, they would need to change this attribute here and that other one there etc.

model/clusters_mgmt/v1/cluster_image_registry_type.model
miguelsorianod: A similar approach is what's implemented in master currently. The difference is that instead of having an array we have a free-form map where the key is the id of the operator.

I explain the detail about that here https://github.com/openshift-online/ocm-api-model/pull/1037#discussion_r2092609890

The use of a map instead of an array is *intentional*. The reason is that arrays are not PATCH friendly. JSON Merge Patch doesn’t support adding, removing, or changing array elements (except by explicitly replacing the whole array).

machi1990:
> @miguelsorianod
> 
> > if the new OCP version stops using previously existing core operators the end user would need to remove them when performing the patch
> 
> Not strictly true. The identity would be unnecessary, but not removing it would not affect cluster operation/upgrade. In either case, this seems orthogonal to the shape of the API, the user behaviour would exist in either case.
> 
> @machi1990
> 
> > I am for option 2; which is feature/functionality centric.
> 
> Yes. I personally like the ergonomics of this approach. However, we need to define it very carefully, since based on @miguelsorianod's comment, it will become awkward for OCP components that require identities and are supported on more than one cloud/platform. That alone is enough to tip me in the other direction.
> 
> Can you update this PR to see how it would look in this case?

@vkareh @miguelsorianod I've pushed https://github.com/openshift-online/ocm-api-model/pull/1037/commits/2271d419a8e7a76aedcea068f49bbfe9c2962ee3

This gives us a model like

```
image_registry:
  state: enabled/disabled
  authentication:
    azure:
      managed_identity:
        controlPlaneIdentity:
        dataPlaneIdentity:
    aws:
      sts:
        controlPlaneIamRole:
        dataPlaneIamRole:
```

model/clusters_mgmt/v1/cluster_image_registry_type.model
machi1990: I've pushed an update on this in line with what Miguel was suggesting above.

deads2k:
The `RequiredOperatorsAuthentication` will not age well as OCP and HCM evolve.  An inability to handle product and service evolution over time blocks the API change until it is resolved.  Adding a hold to prevent accidental merged until the upgrade/evolution problem is resolved.

/hold

deads2k:
> >    I am for option 2; which is feature/functionality centric.
> Yes. I personally like the ergonomics of this approach. However, we need to define it very carefully, since based on @miguelsorianod's comment, it will become awkward for OCP components that require identities and are supported on more than one cloud/platform. That alone is enough to tip me in the other direction.

Thinking through adding additional identities, a choice of adding them per-component requires
1. components remain stable over time and don't move.  This isn't true for OCP, we move parts of components all the time.
2. every time a component is added that needs an identity, addition validation wiring must be added to suit
3. any status reporting code built to ensure the validity of identity post-API call also needs to be updated
4. every time a component is added, even if it has no other configuration, it must have an API change.
5. there is no mechanism to remove a component when it is EOL'd because that produces an incompatibility that makes moving to new schemas more difficult

Can it be done? Yes.  The con-list is notably longer than listed above though.

model/clusters_mgmt/v1/cluster_image_registry_type.model
```diff
@@ -0,0 +1,42 @@
...
+
+	// authentication mechanism to the underlying cloud
+	Authentication OperatorAuthentication
+}
+
+// A union type of cloud authentication mechanism.
+struct OperatorAuthentication {
+	Aws AwsOperatorAuthentication
+	Azure AzureOperatorAuthentication
+}
```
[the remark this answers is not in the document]
machi1990: If I understand correctly "if we are to go with option(2)", the way you'll address the cloud concern is by placing per-component knobs in the top level cloud i.e 
```json
// the cluster spec 
{
...

"azure": {
    "image_registry": {
      ... // image registry knobs
    }
} 
}

```

The reason is that it'll be make it easier for future split of the services - the world where aro-hcp has its own CS separate from rosa-hcp?

machi1990:
Thanks @deads2k for chiming in and additional insights.

> components remain stable over time and don't move.  This isn't true for OCP, we move parts of components all the time.

The per-component API won't have the operator name in it. It'll be a functionality. If the operator name in OCP changes or moves, as long as the desired functionality is there and can be leveraged, I don't see the move having any effect to the API.

>  every time a component is added that needs an identity, addition validation wiring must be added to suit

This can be true as well with a centralised list for the case where an optional identity is added. e.g what do you do when the identity is provided but the feature not enabled? What do you when the identity is not provided when the feature is enabled? 

> any status reporting code built to ensure the validity of identity post-API call also needs to be updated

Same reasoning as above. With the centralised list of identities,  if it is an optional identity, you'll need to report it's status conditionally depending on the feature being enabled or not.

> 4. every time a component is added, even if it has no other configuration, it must have an API change.

True. Taking care of these kinds of issues  was one of the motivations with using a `map` struct originally for identities; the problem that both the RP and CS wanted to solve. However, with some functionality being optional in consequence their operators; we noticed that if we stick with this approach, for a given functionality, we're to have dispersed knobs related to that functionality everywhere - which has its cons. 

> 5. there is no mechanism to remove a component when it is EOL'd because that produces an incompatibility that makes moving to new schemas more difficult

An API deprecation policy is what will address this best. 
However, I think that this con is also applicable to the centralised list approach when the EOL'd component had extra knobs that are defined somewhere in the cluster's spec.

deads2k:
> > 9\. there is no mechanism to remove a component when it is EOL'd because that produces an incompatibility that makes moving to new schemas more difficult
> 
> An API deprecation policy is what will address this best. However, I think that this con is also applicable to the centralised list approach when the EOL'd component had extra knobs that are defined somewhere in the cluster's spec.

While I can appreciate that the current API compatibility is treated fast and loose, with a map (or other arbitrary list), no schema change is required to be communicated to clients and have all clients update to serialize properly.  The wider spread the clients, the harder the required shifts.



> > 7\. every time a component is added, even if it has no other configuration, it must have an API change.
> 
> True. Taking care of these kinds of issues was one of the motivations with using a `map` struct originally for identities `cluster.azure.operators_authentication.managed_identities` field; the problem that both the RP and CS wanted to solve. However, with some functionality being optional in consequence their operators; we noticed that if we stick with this approach, for a given functionality, we're to have dispersed knobs related to that functionality everywhere - which has its cons.


The only listed con is "Harder to distinguish visually which identities are required".  Making the change looks a lot like, "the grass must be greener over there". Think carefully about motivations before taking a plunge with more downsides than the current technique.

deads2k:
> Even though a PR is a good place where to leave comments, have them tracked and have discussions around them, the fact that we are iterating on the custom ocm api model definitions makes hard to understand how would that _actually_ end up looking: As in how the json payload will look like, the openapi definition, the go types, ... . None of those are tracked in the repository and therefore neither in the PR.

Good points. Have a look at
1. https://github.com/openshift-online/ocm-api-metamodel/pull/219
2. https://github.com/openshift-online/ocm-api-model/pull/1024
3. https://github.com/openshift-online/ocm-sdk-go/pull/1037

Which do that coalescing and move the openapi and json types into this repo, while using type aliases to avoid causing problems for ocm-sdk-go consumers.  I'm anticipating that landing early this summer.

model/clusters_mgmt/v1/cluster_image_registry_type.model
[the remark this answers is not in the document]
deads2k: Yes for both what I intend to express in the API and the reasoning behind it.  I think our experience with trying to have a unified API has aged not-so-well and a we should be willing to reflect that in our API and code.

vkareh:
Thanks for this good discussion, folks!

Some things I can see coming from this:

1. Using `{component: {aws: {}, azure: {}}}` will get cumbersome fast. In fact, it already is cumbersome to think/type in that dimension. If we go the way of _de-centralized_ identities, the way I would split it would look more like
```json
{
  "azure": {
    "component": {
      "identity": "..."
    }
  },
  "component": {
    "knob": "..."
  }
}
```
It separates product functionality from cloud requirement.

2. Using the `required*` prefix in the attribute is a hard no. It breaks both options 1 and 2, breaks future-proofing of the API, and confuses the heck out of me when thinking of "optional components that have required identities". :stuck_out_tongue_winking_eye: 

3. Whether identities are part of a required or for an optional component doesn't matter, since a component is either enabled or not, and the identity is therefore required for that component. Therefore, there is no distinction between those identities and so should either co-exist in the same struct, or under each discrete component (caveat: see 1. above).

4. Based on 1. and 3. above, I'm leaning towards co-locating all identities in the same location. Therefore, I would like to see how this API would look like. @machi1990 can you update the API to reflect this, or at least post a sample snippet? That is, unless we're at the point that options 1 or 2 in my original comment start looking very much the same in light of my comment 1. above?

machi1990:
> > > 9. there is no mechanism to remove a component when it is EOL'd because that produces an incompatibility that makes moving to new schemas more difficult
> > 
> > 
> > An API deprecation policy is what will address this best. However, I think that this con is also applicable to the centralised list approach when the EOL'd component had extra knobs that are defined somewhere in the cluster's spec.
> 
> While I can appreciate that the current API compatibility is treated fast and loose, with a map (or other arbitrary list), no schema change is required to be communicated to clients and have all clients update to serialize properly. The wider spread the clients, the harder the required shifts.

 @deads2k The point in my statement (quoted below) 

>  I think that this con is also applicable to the centralised list approach when the EOL'd component had extra knobs that are defined somewhere in the cluster's spec.

is that, if a component is EOL-ed, no schema is required for the centralised approach (option 1) if and only if the component didn't have extra configurations i.e the only configurations for it were the identity information. 
Consider this sample schema that mimics option1;
```
{

"azure": {
   "operators_authentication": {
      "managed_identities": {
         "control_plane_operators": {
             "component1": { // this component has extra configs
                 "resource_id": "foo-bar"
              }
         }
      }
   }
} 
"component1ExtraConfigs": {
  "config1": ...
  "config2" ...
}
}
```

If component1 is EOL'd, the `.azure.operators_authentication.managed_identities_control_plane_operators` map gives you the benefit of no schema change, but you've the change of schema anyway for the case of ".component1ExtraConfigs". So yes, what you said is true only for the case of a component whose configuration is only the identity information.

machi1990:
@vkareh 

> Based on 1. and 3. above, I'm leaning towards co-locating all identities in the same location. Therefore, I would like to see how this API would look like. @machi1990 can you update the API to reflect this, or at least post a sample snippet? That is, unless we're at the point that options 1 or 2 in my original comment start looking very much the same in light of my comment 1. above?

By "I'm leaning towards co-locating all identities in the same location." - that is co-locating only the identity information in one place and the component related information in a different place? i.e pretty much option (1) from https://github.com/openshift-online/ocm-api-model/pull/1037#pullrequestreview-2845016910 

If so, the snippet of that API would look like

```
{

"azure": {
   "operators_authentication": {
      "managed_identities": {
         "control_plane_operators": {
             "image-registry": {
                 "resource_id": "foo-bar"
              }
              .... extra control plane components identities
         }, 
         "data_plane_operators": {
           "image-registry": {
             "resource_id": "bazinga"
           }
           ... extra data plane components identities
         }
      }
   }
} ,
"image-registry": {
  "state": "enabled"
}
}
```

The snippet assumes that we'll get rid of the `.capabilities` section for enablement/disablement of components. If we stick with `.capabilities`, the snippet would look like

```
{

"azure": {
   "operators_authentication": {
      "managed_identities": {
         "control_plane_operators": {
             "image-registry": {
                 "resource_id": "foo-bar"
              }
              .... extra control plane components identities
         }, 
         "data_plane_operators": {
           "image-registry": {
             "resource_id": "bazinga"
           }
           ... extra data plane components identities
         }
      }
   }
}, 
"capabilities": {
  "disabled": []
}
}
```

miguelsorianod:
>   Using {component: {aws: {}, azure: {}}} will get cumbersome fast. In fact, it already is cumbersome to think/type in that dimension. If we go the way of de-centralized identities, the way I would split it would look more like

Doesn't the same apply to the other way around? With the other way around you will have:

```
{
  aws:
    component1:
    componentN:
 azure:
  component1
  componentN
}
```
which is equivalent



> Using the required* prefix in the attribute is a hard no. It breaks both options 1 and 2, breaks future-proofing of the API, and confuses the heck out of me when thinking of "optional components that have required identities". 😜

Using required* as the name is *not* what breaks option 2. What breaks option 2 is the fact that we have a section that is used to designate a set of operators that are required. The aspect that breaks it is the existence of that section where over time some elements in it do not apply anymore. What breaks is the existence of the section and not the name chosen for it. Do you agree with that assessment ? If so, therefore the discussion there with that option is not about the name, it is about finding an alternative that does not leverage that section

model/clusters_mgmt/v1/cluster_image_registry_type.model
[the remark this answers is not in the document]
machi1990: I see and thanks for clarifying. That's approach we could take and it has been considered as well. This is what has been commented as a possible alternatives in the PR's description and left out as a comment as well in the first commit of the PR https://github.com/openshift-online/ocm-api-model/blob/2271d419a8e7a76aedcea068f49bbfe9c2962ee3/model/clusters_mgmt/v1/cluster_type.model#L254

mbarnes:
From an outsider perspective, it seems to me like the central problem here is how to express the **_role_** of an identity in the API.  Each OpenShift version has a discrete set of roles for identities; some required, some not.

I had an alternate idea to the above proposals.  Bear with me because it's not fully baked yet.

Instead of trying to cram this all into the cluster API, what if these roles were exposed as a new set of pre-defined top-level endpoints?  I'll call them "identity roles".  An identity role and its endpoint might look something like:

```
GET /api/clusters_mgmt/v1/identity_roles/{identity_role_id}
{
    "required": true/false,
    "cloud_providers": [
        CloudProviderLink,
        ...
     ],
     "cluster_plane": "control"/"data",
     "purpose": operator or capability name (whatever we decide on)
}
```

Individual identity roles are static.  As new OpenShift versions introduce new operators/functionalities that require identities, Cluster Service could publish new identity roles.  Similarly, if a previously required identity role became optional in later versions (or vice versa), this would entail a new identity role.

The schema for [/api/clusters_mgmt/v1/versions/{version_id}](https://api.openshift.com/#/default/get_api_clusters_mgmt_v1_versions__version_id_) could then be extended to return a list of identity roles appropriate for that version:

```
GET /api/clusters_mgmt/v1/versions/openshift-v4.18.1
{
   ...
   "identity_roles": [
      IdentityRoleLink,
      ...
   }
}
```

Then when a user POSTs a cluster manifest, the `"managed_identities"` section could just be a list (or map) of cluster role ID and resource ID pairings.

```
"azure": {
   "operators_authentication": {
      "managed_identities": [
         {
            "role": "identity-role-1",
            "resource_id": "foo-bar"
         },
         {
            "role": "identity-role-2",
            "resource_id": "bazinga"
         },
         ...
      ]
   }
}, 
```

I'm not sure if identity role IDs should be meaningful names or random like cluster IDs.  I'm leaning toward random, just because naming is hard.  It would require some effort on the user's part to look up identity roles for a given OpenShift version, but that level of effort seems inescapable whatever we settle on.

Maybe there's a seed of a workable idea here.  I'm not as familiar with the design constraints on this API as the rest of you.  I'm just trying to break our thinking out of the box we seem to be stuck in.

vkareh:
@miguelsorianod 
> On top of that, now you will have components that do not have cloud specific configuration and you will end up with something like [...]

Right. That's why my original suggestion called for component-based grouping, but having cloud info underneath is awkward.

However, we need to clarify that identities are for a piece of software (mostly operators in this case) to be allowed to perform a cloud operation on behalf of the user, whereas the component configuration focuses on defining the behaviour of such a component (whether it be an operator or not). So even though they are tied, the component configuration with the identity needed for it to work, they are separate concerns and so it's okay to separate those.

@machi1990 
> If component1 is EOL'd, the `.azure.operators_authentication.managed_identities_control_plane_operators` map gives you the benefit of no schema change, but you've the change of schema anyway for the case of ".component1ExtraConfigs". So yes, what you said is true only for the case of a component whose configuration is only the identity information.

Why would we be okay having a map of arbitrary attributes inside one struct but not on the other?

@mbarnes
Your suggestion is valid and we could explore what that looks like, but you'd still end up with the same structure in the cluster payload anyway, just referencing a identity resource in CS rather than the OCP component that relies on said identity. So for the purpose of this PR it probably doesn't make much of a difference, I think.

vkareh:
@machi1990 
> If so, the snippet of that API would look like [...]

Right, that's closer to what I had in mind, but I would remove some indirections, though:
```json
{
  "azure": {
    "managed_identities": {
      "control_plane": {
        "image_registry_identity_resource_id": "...",
        "some_component_required_identity_resource_id": "...",
        "another_component_optional_identity_resource_id": "...",
        "etc..."
      },
      "data_plane": {
        "image_registry_identity_resource_id": "...",
        "some_component_required_identity_resource_id": "...",
        "another_component_optional_identity_resource_id": "...",
        "etc..."
      }
    }
  },
  "image_registry": {
    "state": "enabled",
    "knob1": "...",
    "knob2": "..."
  },
  "another_component_config": {
    "foo": "bar"
  }
}
```
Since the struct is exclusively for managed identities (whether used by operators, required or not), there's no need for deep nesting, as component knobs don't live there. The data is the same, but with fewer things to move around if/when identity requirements change. You also get all control plane identities in one location, and all data plane identities in a sibling location.

vkareh:
Let's make sure we remove the `"required*"` attribute and any attribute or indirection that would lock us from evolving the API correctly. Adding a hold to prevent merging in the meantime.

/hold

mbarnes:
> Your suggestion is valid and we could explore what that looks like, but you'd still end up with the same structure in the cluster payload anyway, just referencing a identity resource in CS rather than the OCP component that relies on said identity. So for the purpose of this PR it probably doesn't make much of a difference, I think.

I think you might have misunderstood.  Or maybe not and you're seeing issues I've missed.  I guess I still see benefit in my suggestion so let me try to clarify once more here and I'll let it go.

What I'm suggesting is to split out the various ways OCP uses identities into informational, per-component endpoints that live _outside of the cluster API_, like the way the `version.id` cluster field is a reference to the `/api/clusters_mgmt/v1/versions/{version_id}` endpoint, also outside of the cluster API.

Just to take an example of an identity a user might provide:
```
{
    "role": "identity-role-1",
    "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity"
},
```
This associates the user's identity `"myIdentity"` to a predefined identity role endpoint (`/api/clusters_mgmt/v1/identity_roles/identity-role-1`) that specifies:
- What OCP component the identity is used for.
- Whether or not the OCP component is _required_.
- What cloud providers the OCP component applies to.
- Whether the identity is for the control plane or data plane.

And to be clear, Cluster Service would be the one providing these endpoints to cover all supported OCP versions.

The benefits I see are:
- All the identity resource IDs are co-located in the cluster API, which seems to be the consensus here.
- It future-proofs the cluster schema by moving the names of the OCP components that need identities to a separate endpoint.  Over time new OCP components can be introduced, others EOL'd, previously required components can become optional, components that were only on AWS can come to Azure (or vice versa), etc.  All without forcing a schema change.  Cluster Service just adds new identity role endpoints.

miguelsorianod:
> @machi1990
> 
> > If so, the snippet of that API would look like [...]
> 
> Right, that's closer to what I had in mind, but I would remove some indirections, though:
> 
> ```json
> {
>   "azure": {
>     "managed_identities": {
>       "control_plane": {
>         "image_registry_identity_resource_id": "...",
>         "some_component_required_identity_resource_id": "...",
>         "another_component_optional_identity_resource_id": "...",
>         "etc..."
>       },
>       "data_plane": {
>         "image_registry_identity_resource_id": "...",
>         "some_component_required_identity_resource_id": "...",
>         "another_component_optional_identity_resource_id": "...",
>         "etc..."
>       }
>     }
>   },
>   "image_registry": {
>     "state": "enabled",
>     "knob1": "...",
>     "knob2": "..."
>   },
>   "another_component_config": {
>     "foo": "bar"
>   }
> }
> ```
> 
> Since the struct is exclusively for managed identities (whether used by operators, required or not), there's no need for deep nesting, as component knobs don't live there. The data is the same, but with fewer things to move around if/when identity requirements change. You also get all control plane identities in one location, and all data plane identities in a sibling location.

A discussion around that additional nesting section happened a while ago and it was decided to go forward with it. You were involved in that conversation if I recall correctly.
To refresh about it, control plane identities are currently in a single location already, same for data plane identities, and they are siblings as you suggested.
The only difference is that in master we have the additional intermediate layer operators_authentication. That was intentional and I would like to keep it that way for the following reasons:
* It semantically describes better that it is related to operators authentication. Having a direct managed_identities section is less clear to me. Managed identities for what
* It allows future expansion at the operators_authentication level where there could be other configurations related to operators authentication and/or allowing other kinds of authentication other than managed identities

machi1990:
> > Your suggestion is valid and we could explore what that looks like, but you'd still end up with the same structure in the cluster payload anyway, just referencing a identity resource in CS rather than the OCP component that relies on said identity. So for the purpose of this PR it probably doesn't make much of a difference, I think.
> 
> I can't tell from your response if I didn't explain this clearly enough or if I'm just misreading the problem we're trying to solve. If you'll humor me a little longer, let me try to clarify once more and then I'll let it go.
> 
> What I'm suggesting is to split out the various ways OCP uses identities into informational, per-component endpoints that live _outside of the cluster API_, like the way the `version.id` cluster field is a reference to the `/api/clusters_mgmt/v1/versions/{version_id}` endpoint, also outside of the cluster API.
> 
> Just to take an example of an identity a user might provide:
> 
> ```
> {
>     "role": "identity-role-1",
>     "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity"
> },
> ```
> 
> This associates the user's identity `"myIdentity"` to a predefined identity role endpoint (`/api/clusters_mgmt/v1/identity_roles/identity-role-1`) that specifies:
> * What OCP component the identity is used for.
> * Whether or not the OCP component is _required_.
> * What cloud providers the OCP component applies to.
> * Whether the identity is for the control plane or data plane.
> 
> And to be clear, Cluster Service would be the one providing these endpoints to cover all supported OCP versions.
> 
> The benefits I see are:
> 
> * All the identity resource IDs are still co-located in the cluster API, which seems to be the consensus here.
> * It future-proofs the cluster schema by moving the names of the OCP components that need identities to a separate endpoint.  Over time new OCP components can be introduced, others EOL'd, previously required components can become optional, components that were only on AWS can come to Azure (or vice versa), etc.  All without forcing a schema change.  Cluster Service just adds new identity role endpoints.

@mbarnes Thank you for chiming in. In this model, how are the managed identities specified in the cluster payload when creating a cluster? Or that association (this cluster needs this and that managed identities info for `foo` component, and that identity for `bar` component)  is done somewhere else; where and how is that done?

miguelsorianod:
> @vkareh @deads2k @miguelsorianod @mbarnes please see the example schema below of what I thought option (2) could look like with the two challenges stated out in the PR's description resolved.
> 
> Cloud specific components are in the cloud specific part of the cluster's spec. Generic component/features resides in the top level cluster's spec, the configuration of it are separated par cloud; a union type has been chosen here.
> 
> In this schema, there is no map/list of identities. But rather per-component knobs.
> 
> **NOTE**: Naming are not finalised and open to proposition or questions about what concepts they are referring to. **NOTE**: I've not updated the model to include this change. So please, look at the schema below.
> 
> ```json
> {
>  "azure": {
>      "subscription_id": "sub-id",
>      "resource_name": "resource-name",
>      ....
> 
>      "azure_cloud_api_authentication": { // a union type defining different azure cloud authentication mechanim.
>        "managed_identity": { // this section identities the MSI RP identity url and the service managed identity.
>             "msi_data_plane_identity_url": "https://dummyhost.identity.azure.net/", // this is used by all control plane components managed identities
>             "service_managed_identity": {
>                 "resource_id": "foo-bar"
>             }
>        }
>       }, 
>       "cluster_api_azure": { // all azure specific components will be configured under the azure section like here
>         "cloud_api_authentication": {
>             "managed_identity": {
>                 "resource_id": "cluster-api-azure-resource-id"
>             }
>         }
>       }
>       "another_azure_specific_functionality_that_requires_mi": {
>         "cloud_api_authentication": {
>             "managed_identity": {
>                 "resource_id": "some-value"
>             }
>         }
>       }
> },
> // below are cloud agnostic functionalities that shows how option(2) may look like
> "image_registry": {
>     "state": "enabled",
>     "cloud_api_authentication": { // a union type containing different authentication mechanism depending on the cluster's cloud
>         "azure": {
>             "managed_identity": { // an example with both data plane and control plane specified
>                 "control_plane_identity": {
>                      "resource_id": "foo-bar"
>                 },
>                 "data_plane_identity": {
>                     "resource_id": "foo-bar"
>                 }
>             }
>         },
>         "aws": { // this shows a possibility where we chose to inline the authentication mechanism
>             "sts_control_plane_iam_role": "some-arn-role",
>             "sts_data_plane_iam_role": "some-other-role"
>         }
>      }
> },
> "ingress": { // this is an always required component. No need to .state=enabled/disabled. 
>     "cloud_api_authentication": { // a union type containing different cloud authentication mechanism depending on the cluster's cloud
>         "azure": {
>             "managed_identity": { // an example where only a single identity is needed e.g the control plane one
>                 "resource_id": "foo-bar"
>             }
>         },
>         "aws": {
>             "iam_role": "some-role"
>         }
>     }
> },
> "another_generic_functionality": {
>     "cloud_api_authentication": {
>         "managed_identity": {
>             "resource_id": "some-value"
>         }
>     }
> }
>     ...other components here
> }
> ```
> 
> What are your thoughts here?

+1. As I discussed with you that's how we think this design would be.
A remaining part, which you mentioned already, is the naming of some of the concepts. I am still unclear about some of them. I also wonder about whether the top-level `azure.cloud_api_authentication` naming/structure too.

machi1990:
@mbarnes 
> The user still provides the resource ID for the identity. But instead of trying to declare all the cluster components that take identities directly in the cluster API, those components would instead be split out into a separate endpoint I've been calling "identity roles". Cluster Service would provide these, perhaps from a config map on the service cluster.
> 
> An example of an identity role might look like this:
> 
> ```
> GET /api/clusters_mgmt/v1/identity_roles/identity-role-1
> {
>   "kind": "IdentityRole",
>   "id": "identity-role-1",
>   "href": "/api/clusters_mgmt/v1/identity_roles/identity-role-1",
>   "required": true,
>   "cloud_providers": [
>     {
>       "kind": "CloudProviderLink",
>       "id": "aws",
>       "href": "/api/clusters_mgmt/v1/cloud_providers/aws"
>     },
>     {
>       "kind": "CloudProviderLink",
>       "id": "azure",
>       "href": "/api/clusters_mgmt/v1/cloud_providers/azure"
>     }
>   ],
>   "cluster_plane": "control",  // i.e. control plane... needs a better name?
>   "cluster_component": "image_registry",
> }
> ```
> 
> How would a user ever discover these? 

I see. This is more or less the "managed identity requirements" information endpoint that we are working on in the 1095 strat. There will be an endpoint that will provide this sort of information to the end user. In the strat there is a design doc capturing the proposal made for both the RP and CS and we can talk about that separately.

> So iterating on the example in [this recent comment](https://github.com/openshift-online/ocm-api-model/pull/1037#issuecomment-2887655129), the user would just pair an identity resource ID with an identity role ID for the OpenShift version being installed. I'm unsure whether these pairing would best be expressed as a list or a map. I showed the list option in a previous comment so here's what a map might look like.
> 
> ```
> {
>   "azure": {
>     "managed_identities": {
>         "identity-role-1": {
>             "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity1",
>         },
>         "identity-role-2": {
>             "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity2",
>         },
>         ...
>     ]
>   },
>   "image_registry": {
>     "state": "enabled",
>     "knob1": "...",
>     "knob2": "..."
>   },
>   "another_component_config": {
>     "foo": "bar"
>   }
> }
> ```
> 
> Cluster Service would then validate that -- for the current cloud provider and OpenShift version -- the provided list (or map) satisfies all the _required_ identity roles, that there are no unsupported identity roles (i.e. ones not listed at the version endpoint), etc.

> The benefit (as I see it) is it allows the set of cluster components that take identities to gradually change from version to version without disrupting the cluster schema at all. A particular identity role would be immutable forever, but it could be phased out and replaced by new ones over time.
> 
> Note this doesn't address component-specific knobs at all. In this scheme, the knobs are a separate concern.

While looking at this example, if I understood it correctly, the model proposed here isn't so different from https://github.com/openshift-online/ocm-api-model/pull/1037#issuecomment-2887655129 (option 1) except that instead of using a component name/operator name which are currently using.

mbarnes:
> This is more or less the "managed identity requirements" information endpoint that we are working on in the 1095 strat. There will be an endpoint that will provide this sort of information to the end user. In the strat there is a design doc capturing the proposal made for both the RP and CS and we can talk about that separately.

Yeah, Miguel pointed out to me the parallels with 1095.  I'm not sure if the currently proposed CS API would be compatible with the scheme I'm suggesting.  You would kind of need a way to reference _into_ the `managed_identity_requirements` response data to point at a particular operator/component name.   But there's definitely a ton of common ground.

vkareh:
@miguelsorianod 
> It allows future expansion at the operators_authentication level where there could be other configurations related to operators authentication and/or allowing other kinds of authentication other than managed identities

I suspect that if we use something other than managed identities in the future, it's such a large change in our architecture that there will be plenty of other API-breaking changes anyway. Using that attribute is not wrong, but likely unnecessary.

@machi1990
> please see the example schema below [...]

This approach becomes problematic for long-term maintenance of the API. Having each piece of functionality itself be cloud-aware will potentially slow down feature development and cause conflict across the various services/offerings. Not to mention that it will really mess things up if end up doing the things that led to adding the `operators_authentication` attribute in the first place.
You'd end up with something like:
```json
"image_registry": {
  "cloud_api_authentication": {
    "azure": {
      "managed_identity": {
        "control_plane_identity": {
          "resource_id": "..."
        },
        "data_plane_identity": {
          "resource_id": "..."
        }
      },
      "app_config": {
        "control_plane_identity": {
          "client_id": "...",
          "client_secret": "...",
          "tenant_id": "..."
        },
        "data_plane_identity": {
          "client_id": "...",
          "client_secret": "...",
          "tenant_id": "..."
        }
      }
    },
    "aws": {
      "sts": {
        "control_plane_identity": {
          "role_arn": ""
        }
        "data_plane_identity": {
          "role_arn": ""
        }
      },
      "iam": {
        "control_plane_identity": {
          "access_key_id": "...",
          "secret_access_key": "..."
        },
        "data_plane_identity": {
          "access_key_id": "...",
          "secret_access_key": "..."
        }
      }
    }
  }
},
```
This struct is just for a single operator/component, and it doesn't even have knobs yet.
We should not pursue that path.

@mbarnes
> let users reference the names from the managed identity requirements endpoint instead of baking that information into the cluster schema.

To me the cluster spec would look like the following:
```json
{
  "azure": {
    "managed_identities": {
      "image_registry_identity_ref": "/api/clusters_mgmt/v1/identity_roles/identity-role-1",
      "some_other_component_identity_ref": "/api/clusters_mgmt/v1/identity_roles/identity-role-1",
    }
  },
  "image_registry": {
    "state": "enabled"
  }
}
```
And then the identity resource would look like:
```json
{
  "component": "image_registry",
  "cloud_provider": "azure",
  "azure": {
    "control_plane": {
      "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity1"
    },
    "data_plane": {
      "resource_id": "/subscriptions/.../Microsoft.ManagedIdentity/userAssignedIdentities/myIdentity1"
    }
  }
}
```
Honestly this could possibly be the cleanest version of the cluster spec in terms of identity, since the mess can be contained in a separate, potentially evolving, resource, and the identities are all grouped in the same location within the cluster spec.

miguelsorianod:
> I suspect that if we use something other than managed identities in the future, it's such a large change in our architecture that there will be plenty of other API-breaking changes anyway. Using that attribute is not wrong, but likely unnecessary.

We don't know how it will evolve, but it still keeps open the possibility as we do with other cases like when we have a functionality where only enablement is available where we decide to put it within a struct. As I mentioned it also allows to add other operators authentication configuration not coupled to managed identities or any other authentication method. And it adds more semantics to describe the purpose of managed identities.

miguelsorianod:
> This approach becomes problematic for long-term maintenance of the API. Having each piece of functionality itself be cloud-aware will potentially slow down feature development and cause conflict across the various services/offerings

Could you expand on this? I am trying to understand why. If a functionality has a dependency on one or multiple cloud providers to be able to work, it will need to configuration associated to it, no matter how you organize the API.

miguelsorianod:
>  You'd end up with something like:

What is the issue you detect with that? the length of the json? assuming the information is in the Cluster payload, you will need to have that information no matter how you organize it. You might save some intermediate sections if you put it all together but then the concern about having disjointed configurations kicks in anyway, which is not being addressed.

> and the identities are all grouped in the same location within the cluster spec.

The approach mentioned and the original approach #1 have the identities are grouped in the same location within the cluster spec. The new approach you mentioned does not have any difference in that regard.

I am going to attempt to analyze the approach, which I understand builds on Matthew's comment:

A difference is that the Cluster json payload would be shorter.

Another difference is that now the user will reference other CS API resources URLs instead of having the information inline. This also implies the users will perform N additional CS API calls before creating a Cluster.

Something that indicates that the solution might not work, or needs more refinement is that the identity roles are being defined at a *global* level /api/clusters_mgmt/v1/identity_roles/. There, a "component" (operator) is defined as well as the identities associated to it. However, each Cluster has to have their own unique set of managed identities associated to the components it uses. The solution does not seem to work as you have a set of identities per component.
If a new approach is being proposed I would like to see a clear definition on how it would work: as in, defined resources, endpoints, API attributes, how would cloud providers influence it, the user interaction flow, an analysis of the implications and so on.

machi1990:
+1 to @miguelsorianod assessment above. 

> This approach becomes problematic for long-term maintenance of the API. Having each piece of functionality itself be cloud-aware will potentially slow down feature development and cause conflict across the various services/offerings. 

@vkareh on top of what @miguelsorianod mentioned here https://github.com/openshift-online/ocm-api-model/pull/1037#issuecomment-2897083324 I would love to see an example schema on how any proposal built on option (1)   i.e the centralised approach, plans to address the eventual possibility of a functionality/component needing other (different than managed identities) cloud specific configurations.

tzvatot:
> please see the example schema below of what I thought option (2) could look like

My main concern with this option is it does not scale - as every cluster level feature that has different implementation details (which require "knobs") will force us to create a platform section (i.e. cloud provider) underneath.

vkareh:
@machi1990 no, that doesn't match what Elad and I discussed. Here's why we're choosing option 1 above:
- **identities under cloud struct**: Since identities are a cloud-specific concern, the API will group them under the corresponding cloud struct. This ensures the API keeps all cloud-specific configuration together and makes the overall API portable.
- **identities grouped together**: Since identities are under the cloud struct, but the functionality they enable is not strictly cloud-specific, the API will group all identities together to keep the API self-consistent (rather than splitting on identities for cloud-specific functionality vs identities for cluster-wide functionality - often the line between these blurs and would cause confusion).
- **no separate category between required and optional**: Since we cannot predict what components transition between required and optional, and because identities can be required for optional components to work, the API struct will not make an explicit distinction between required and optional components. That is left up to the API validation/preflights/etc to implement as needed.
- **components under the broadest category that they support**: if a component is cluster-wide (e.g. image registry), then the configuration lives outside of the cloud struct. This ensures reusability accross clouds/offerings.

We will proceed with updating the API Guidelines to ensure this is explicit moving forward. Based on this, a possible implementation for this particular API change could look like:
```json
{
  "azure": {
    "operators_authentication": {
      "managed_identities": {
        "control_plane": {
          "image_registry_identity": {
            "resource_id": "..."
          }
          "some_component_identity": {
            "resource_id": "..."
          }
          "another_component_identity": {
            "resource_id": "..."
          }
        },
        "data_plane": {
          "image_registry_identity": {
            "resource_id": "..."
          }
          "some_component_identity": {
            "resource_id": "..."
          }
          "another_component_identity": {
            "resource_id": "..."
          }
        }
      }
    }
  },
  "image_registry": {
    "state": "enabled",
    "knob1": "...",
    "knob2": "..."
  },
  "another_component_config": {
    "foo": "bar"
  }
}
```
With regards to the `operators_authentication` and nesting of `resource_id`, I see it as unnecessary cruft but it doesn't technically deviate from the guidelines outlined above, so I'll leave it up to the ARO team to determine if they want to maintain those indirections or not.

@machi1990 can you update the PR to reflect these API changes?

machi1990:
@vkareh @tzvatot 

I'll try to summarise my understanding of your proposal

There are two proposal;

- (1): We keep the current API design around managed identities and use it for both required and optional components
- (2): We remove the notion of the ocp capabilities from our API and instead move the enablement/disablement of a component will be located along the rest of "extra" the knobs for that particular component; extra knobs here means different than managed identities

> components under the broadest category that they support: if a component is cluster-wide (e.g. image registry), then the configuration lives outside of the cloud struct. This ensures reusability accross clouds/offerings.

To this point, I want to make sure that you've seen this https://github.com/openshift-online/ocm-api-model/pull/1037#issuecomment-2897215207 and how you envision the proposal to cater for this? If the ask wasn't clear, I can try to clarify. 
 
> can you update the PR to reflect these API changes?

I am okay to update the API if we confirm that the second decision around removing the ocp capabilities. To this part let me tag along @flavianmissi as the designer of that API for thoughts and @mbarnes for the RP side as it might be a decision we would want to take along as well for the RP.

miguelsorianod:
As discussed through other channels with @vkareh we are in agreement that the set of operator identities of a given type (differentiated by control plane / data plane) will be defined as a free-form map[string, <controlplane/dataplane identity>]. Remarks about this decision are:
* We are not defining it as an array because arrays are not json merge patch friendly. The reasons on why are explained in other comments in this PR
* By having it as a map as opposed of as a struct of N attributes, if a new identity needs to be defined or removed, the API scheme definition will not change

The key, which will be a string, will be a key we define that indicates the operator name associated to the identity.

You can see how it is implemented in the master branch already @vkareh, in case you want to take a look at the defined data types and attributes.

vkareh:
To follow up on an example of a feature that contains cloud-specific configurations, we have etcd encryption. Today the model looks like:
```json
{
  "etcd_encryption": true,
  "aws": {
    "etcd_encryption": {
      "kms_key_arn": "..."
    }
  },
  "gcp_encryption_key": {
    "kms_key_service_account": "..."
    "key_location": "...",
    "key_name": "...",
    "key_ring": "..."
  }
}
```
Under the guidelines I outlined, this is largely correct (even though it should be nested in the case of GCP and ideally we refrain from using a boolean for the top-level config). If we follow the guidelines, the API should look like:
```json
{
  "etcd_encryption": {
    "state": "enabled"
  },
  "aws": {
    "etcd_encryption": {
      "kms_key_arn": "..."
    }
  },
  "azure": {
    "managed_identities": {
      "etcd_encryption": {
        "resource_id": "..."
      }
    }
  },
  "gcp": {
    "etcd_encryption": {
      "kms_key_service_account": "..."
      "key_location": "...",
      "key_name": "...",
      "key_ring": "..."
    }
  }
}
```

miguelsorianod:
Ok so if you think that's the correct design, and you agree on Manyanda's example, https://github.com/openshift-online/ocm-api-model/pull/1037#issuecomment-2911386942 **including key names for identities**.

Then I think the way forward with this MR is to close it, because that's the current design that it is already in the master branch regarding the managed identities area.

Can you confirm that we are ok with this way forward? If that's the case, we'll close the MR.

machi1990:
Below is the schema of what we've in main. The schema captures the configuration of etc encryption for 3 clouds we support. 

For the aro-hcp case, the `cluster.azure.operators_authentication.managed_identities` contains two subsections which are `control_plane_operators` and `data_plane_operators`. These sections are represented by a Map struct, the key being the operator name.

```json
{
  "etcd_encryption": {
    "state": "enabled"
  },
  "aws": { 
     "sts": {
        "operator_iam_roles": [
          .... {
            "name": "kms-provider",
            "role_arn": "kms-role-arn",
            ....
          }
       ]
     },
    "etcd_encryption": {
      "kms_key_arn": "..."
    }
  },
  "azure": {
    "etcd_encryption": {
        "foo_bar_ksm_key": {
          ...Azure KV related info
        }
      },
      "operators_authentication": {
           "managed_identities": {
                "control_plane_operators": { // this is a Map struct with key being the operator
                   ...
                  "kms": { // here the key is the operator name
                      "resource_id": ....
                    }
                }, 
                "data_plane_operators": { // this is a Map struct with key being the operator name
                  ....
                }
             }
       }
  },
  "gcp": {
    "etcd_encryption": {
      "kms_key_service_account": "..."
      "key_location": "...",
      "key_name": "...",
      "key_ring": "..."
    }
  }
}
```

@vkareh @tzvatot if you agree with that, I'll close the PR.

tzvatot:
> 

> ```json
> "kms": { // here the key is the operator name
> ```
Is the operator name "kms"?

tzvatot:
> The schema captures the configuration of etc encryption for 3 clouds we support.

Schema looks good to me. @vkareh ?

machi1990:
> > 
> 
> > ```json
> > "kms": { // here the key is the operator name
> > ```
> 
> Is the operator name "kms"?

Yes. The name of the key in the payload will be whatever is in this config file https://github.com/Azure/ARO-HCP/blob/main/cluster-service/deploy/templates/azure-operators-managed-identities-config.configmap.yaml#L89 (the key in the config there are the operator names)

PR #1042, opened 2025-05-30
Remove old github workflows that have been replaced by prow

deads2k:
We are now wired to prow for our pre-merge testing and we run the same set of make targets.

https://github.com/openshift/release/blob/master/ci-operator/config/openshift-online/ocm-api-model/openshift-online-ocm-api-model-main.yaml#L18 and https://github.com/openshift-online/ocm-api-model/blob/main/Makefile#L44


vkareh:
@deads2k is there a way to see the output of `verify`/`make check`?
All I see is:
```
INFO[2025-05-30T18:09:40Z] Executing test verify                        
INFO[2025-05-30T18:09:58Z] Ran for 4m16s                                
```
and nothing in between those two

deads2k:
artifiacts in the upper right, then navigate to https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-online_ocm-api-model/1042/pull-ci-openshift-online-ocm-api-model-main-verify/1928513112810459136/artifacts/test/build-log.txt

on success, the build-log isn't displayed on the main page.  We can write someting to make junit.xml if we want to see details on success I guess.

vkareh:
> We can write someting to make junit.xml if we want to see details on success I guess.

No, this is sufficient - I just wanted to make sure we have access to those logs for every test run.

vkareh (approved):
/lgtm

PR #1044, opened 2025-06-03
update gh setup-python to latest version

ahitacat:
Cache service was recently decommissioned [[link](https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts)]. 

This makes old versions of action that uses cache to fail (see metamodel action [result](https://github.com/openshift-online/ocm-api-metamodel/actions/runs/15412875097) ). 

According to this [issue](https://github.com/orgs/community/discussions/160793) it is recommended to keep the actions up to date.

>For others seeing this message, please ensure that your packages, and the dependencies of those packages, are all using an up-to-date version of the cache service.

### DIFF
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -31,7 +31,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v5
       with:
         python-version: '3.10'
         cache: 'pip'


nimrodshn:
LGTM :100:

miguelsorianod:
By looking at it, I understand the issue where we fall is: 

```
Using a third party action (i.e. not actions/cache) or product that uses an actions cache service to perform caching. In this case, you may need to upgrade to the latest version. Examples: [mozilla/sccache](https://github.com/mozilla/sccache), [Mozilla-Actions/sccache-action](https://github.com/Mozilla-Actions/sccache-action), [Docker with GitHub Actions as a caching backend](https://docs.docker.com/build/cache/backends/gha/)
```

And that the `actions/setup-python` version we were referencing was using the decommissioned service. And to fix it we updated to a new version of that dependency which does not use the decommissioned service anymore.

If that's the case, lgtm

ahitacat:
@miguelsorianod you understood correctly :smile: Can you give the `/lgtm` so the bot can merge this?

PR #1046, opened 2025-06-05
OCM-15929 : Add LogType to notify_details_request_type.model

tirthct:
**Changed**
- Add LogType to notify_details_request_type.model


model/accounts_mgmt/v1/notify_details_request_type.model
```diff
@@ -37,4 +37,7 @@ struct NotificationDetailsRequest {
 
 	// Indicates whether the service log is internal only to RH.
 	InternalOnly Boolean
+
+	// Indicates the type of the service log.
+    LogType String
```
davidleerh, line 42: Would you be able to see if the spacing can be corrected here?

davidleerh:
wondering if you can also add a second (not squashed - separate) commit with the update to changes.md :slightly_smiling_face:

model/accounts_mgmt/v1/notify_details_request_type.model
macgregor, line 42: is this actually a string or an enum type erased to a string? not sure if the meta-model supports specifying valid values, but if so we should do that. basically we want as much information here as we would want in openapi docs or a generated client to use the api correctly without needing to go to a secondary source or person for clarification. good descriptions, allowed values, etc.

maybe thats something we come back to in the future in our quest for a single source of truth for API definitions...

tirthct: Ah, fixed the whitespace

tirthct:
> wondering if you can also add a second (not squashed - separate) commit with the update to changes.md 🙂

Ack. Added a new commit for updating changes.md

davidleerh:
would you be able to squash these two commits? :pray: 
image

davidleerh (approved):

PR #1049, opened 2025-06-12
feat: update aro_hcp/v1alpha1 add nodepools to be async

JameelB:
The following changes includes:
- Update of the ocm-api-metamodel to [v0.0.69](https://github.com/openshift-online/ocm-api-metamodel/releases/tag/v0.0.69) (see changes here: https://github.com/openshift-online/ocm-api-metamodel/compare/v0.0.68...v0.0.69).
  - No additional changes were applied when running `make update` and `make verify` with the new metamodel version.
- Update aro_hcp/v1alpha1 node_pools Add method to be async, using `AsyncAdd`, following the feedback from https://github.com/openshift-online/ocm-api-model/pull/1047#issuecomment-2966220903
  - Since CS is already returning 202, this should be reflected in the API definition.
  - ARM uses 202 for POST asynchronous action, back then it was decided to use the same for POST async in CS as well.
  - Also, conceptually, the resource has been accepted but it is not yet fully created, therefore 202 would be more suited as a response.  
  - `make update` and `make verify` was ran to generate the openapi changes.
- Update aro_hcp/v1alpha1 clusters Add and Delete method to be async, using `AsyncAdd` and `AsyncDelete`
  - Since CS is already returning 202 for both of these endpoints, this should be reflected in the API definition.


model/aro_hcp/v1alpha1/node_pools_resource.model
```diff
@@ -62,7 +62,8 @@ resource NodePools {
 	}
 
 	// Adds a new node pool to the cluster.
-	method Add {
+    @go(name="Add")
+	method AsyncAdd {
```
ahitacat, line 66:
```suggestion
        @go(name="Add")
	method AsyncAdd {
```
Can this be aligned?

JameelB: good spot! update now.

metamodel_generator/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/openshift-online/ocm-api-model/metamodel_generator
 
 go 1.23.0
 
-require github.com/openshift-online/ocm-api-metamodel v0.0.69-0.20250609135152-2defa0ded8b3
+require github.com/openshift-online/ocm-api-metamodel v0.0.69-0.20250612120255-488a7c2d8750
```
miguelsorianod, line 5: This seems a change in behavior.

Before we were creating new releases in the ocm-api-metamodel but now I observe that we are referencing the same tag but with a different commit.

Before we didn't need to specify a timestamp. Rather we created new tags that could be used as a dependency.

Is this intentional?

JameelB: Im not sure why this is the case now, I noticed this in the last update to the metamodel. @ahitacat mentioned we no longer need to do releases for the metamodel, and i'm unsure why. 

Do you know why this is the case @ahitacat, @nimrodshn?

ahitacat: It is not required to do a release. 
But after talking this with @miguelsorianod we think we should keep having the same workflow, and create a new release on the `metamodel`.

JameelB: This has been updated to use v0.0.69

JameelB:
@ahitacat @miguelsorianod this has been updated to bump the version of metamodel to v0.0.69

miguelsorianod (approved):

PR #1050, opened 2025-06-13
[DRAFT] ARO-15052: support cluster credential retrieval for aro_hcp/v1alpha1

JameelB:
h4. What

The design document for this API can be found in  [XCMSTRAT-1135](https://issues.redhat.com/browse/XCMSTRAT-1135).

The proposed endpoints in ARO HCP v1alpha1 is as follows:

```
GET POST DELETE
/clusters/{cluster_id}/break_glass_credentials

GET 
/clusters/{cluster_id}/break_glass_credentials/{break_glass_credential_id}
```

The break glass credentials API for both _clusters_mgmt/v1_ and _aro_hcp/v1alpha1_ should have similar features apart from the following differences:
- Creating and deleting break glass credentials for ARO HCP clusters needs to be done asynchronously. The POST and DELETE endpoints must return a 202 Accepted status code.
- Additional states were added: `issuing`, `revoking` and `denied`

A few additional properties has been added to _clusters_mgmt/v1_ `BreakGlassCredentials` type:
- *CreationTimestamp*: added in order to determine when to clean up 'denied' and 'failed' credentials. It may also be useful for auditing/tracing. 
   
  For existing credentials, these will be set to a default of zero time. These will be omitted in the API output. Moving forward, all break glass credentials under both _clusters_mgmt/v1_ and _aro_hcp/v1alpha1_ will have this property set.
- *StatusDetails*: added to contain additional properties about the break glass credential's state such as 'message'. As there is already an existing `Status` property, we can't align its name with the rest of the other endpoints like Node Pools to capture this information. A few alternative options on naming were considered such as 'State' or 'StatusContext'.

Additional endpoints that we may consider adding:
```
GET
/clusters/{cluster_id}/break_glass_credentials/{break_glass_credential_id}/status
```
- A similar proposal was made for external authentication provider, based on the discussion there, we may want to consider adding this endpoint for break glass credentials as well: https://github.com/openshift-online/ocm-api-model/pull/1047#discussion_r2140398429

Only the openapi files have been generated for now to make it easier to review the changes here.

Resulting changes in the ARO HCP API
image

### DIFF
--- a/model/aro_hcp/v1alpha1/break_glass_credential_resource.model
+++ b/model/aro_hcp/v1alpha1/break_glass_credential_resource.model
@@ -1,0 +1,23 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a specific break glass credential.
+resource BreakGlassCredential {
+	// Retrieves the details of the break glass credential.
+	method Get {
+		out Body BreakGlassCredential
+	}
+}
--- a/model/aro_hcp/v1alpha1/break_glass_credentials_resource.model
+++ b/model/aro_hcp/v1alpha1/break_glass_credentials_resource.model
@@ -1,0 +1,81 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages the break glass credentials of a cluster.
+resource BreakGlassCredentials {
+	// Retrieves the list of break glass credentials.
+	method List {
+		// Index of the requested page, where one corresponds to the first page.
+		in out Page Integer = 1
+
+		// Number of items contained in the returned page.
+		in out Size Integer = 100
+
+		// Search criteria.
+		//
+		// The syntax of this parameter is similar to the syntax of the _where_ clause of a
+		// SQL statement, but using the names of the attributes of the break glass credentials
+		// instead of the names of the columns of a table. For example, in order to retrieve all
+		// the credentials with a specific username and status the following is required:
+		//
+		// ```sql
+		// username='user1' AND status='expired'
+		// ```
+		//
+		// If the parameter isn't provided, or if the value is empty, then all the
+		// break glass credentials that the user has permission to see will be returned.
+		in Search String
+
+		// Order criteria.
+		//
+		// The syntax of this parameter is similar to the syntax of the _order by_ clause of
+		// a SQL statement, but using the names of the attributes of the break glass credentials
+		// instead of the the names of the columns of a table. For example, in order to sort the
+		// credentials descending by identifier the value should be:
+		//
+		// ```sql
+		// id desc
+		// ```
+		//
+		// If the parameter isn't provided, or if the value is empty, then the order of the
+		// results is undefined.
+		in Order String
+
+		// Total number of items of the collection.
+		out Total Integer
+
+		// Retrieved list of break glass credentials.
+		out Items []BreakGlassCredential
+	}
+
+	// Adds a new break glass credential to the cluster.
+	@go(name="Add")
+	method AsyncAdd {
+		// Description of the break glass credential.
+		in out Body BreakGlassCredential
+	}
+
+	// Revokes all the break glass certificates signed by a specific signer.
+	@go(name="Delete")
+	method AsyncDelete {
+	}
+
+	// Reference to the service that manages a specific break glass credential.
+	locator BreakGlassCredential {
+		target BreakGlassCredential
+		variable ID
+	}
+}
--- a/model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
+++ b/model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
@@ -1,0 +1,34 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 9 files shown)


model/clusters_mgmt/v1/break_glass_credential_type.model
```diff
@@ -28,16 +31,22 @@ class BreakGlassCredential {
 	// Status is the status of this credential
 	Status BreakGlassCredentialStatus
 
+    // Additional information about the current state of this credential
+    StatusDetails BreakGlassCredentialStatusDetails
```
machi1990, line 35: np; indentation

model/clusters_mgmt/v1/break_glass_credential_type.model
```diff
@@ -28,16 +31,22 @@ class BreakGlassCredential {
...
 	Created
+    Issuing
 	Issued
 	Expired
 	AwaitingRevocation
+    Revoking
 	Revoked
 	Failed
+    Denied
 }
```
machi1990, line 52: np; indentation

model/aro_hcp/v1alpha1/break_glass_credentials_resource.model
```diff
@@ -0,0 +1,81 @@
...
+
+		// Total number of items of the collection.
+		out Total Integer
+
+		// Retrieved list of break glass credentials.
+		out Items []BreakGlassCredential
+	}
+
+	// Adds a new break glass credential to the cluster.
+    @go(name="Add")
```
machi1990, line 65: np; indentation.

Same below for Delete

model/clusters_mgmt/v1/break_glass_credential_type.model
JameelB: this must be my vscode setting cause it's indented properly there 😂 I'll fix, thanks

model/aro_hcp/v1alpha1/break_glass_credentials_resource.model
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
```
machi1990, line 2:
```suggestion
Copyright (c) 2025 Red Hat, Inc.
```

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -296,6 +296,218 @@
...
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        }
+      },
+      "delete": {
```
machi1990, line 344: same https://github.com/openshift-online/ocm-api-model/pull/1050/files#r2144998351

model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
```diff
@@ -0,0 +1,34 @@
...
+
+// Representation of the status details of a break glass credential
+class BreakGlassCredentialStatusDetails {
+    // The current state of the credential
+    State BreakGlassCredentialState
+    
+    // A descriptive message providing additional context about the current state of
+    // the credential
+    Message String
+}
```
JameelB, line 25: The initial proposal for this property was as follows:
```
"status_details": {
    "message": "cert failed to be issued for x reason"
}
```

However, after reviewing the node pool state, I'm proposing to align this with the [NodePoolStatus](https://github.com/JameelB/ocm-api-model/blob/main/model/clusters_mgmt/v1/node_pool_status_type.model) and [NodePoolState](https://github.com/JameelB/ocm-api-model/blob/main/model/clusters_mgmt/v1/node_pool_state_type.model) types to have consistency as to how the status/state is defined across the API. 

Instead, the new property will be as follows:
```
"status_details": {
    "kind": "string",
    "id": "string",
    "href": "string",
    "message": "string",
    "state": {
         "last_updated_timestamp": "2025-06-13T12:43:34.624Z",
         "value": "issuing"
    }
},
```

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -296,6 +296,218 @@
...
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/BreakGlassCredential"
+              }
+            }
+          }
+        },
+        "responses": {
+          "201": {
```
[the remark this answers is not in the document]
JameelB: yes its async, but I'm waiting on https://github.com/openshift-online/ocm-api-model/pull/1049 to get merged so i can generate the async add method. That PR updates the metamodel that includes support for that. 

You'll notice below that Delete is the same. I'm going to regenerate this once that pr is merged.

openapi/aro_hcp/v1alpha1/openapi.json
JameelB: see comment above

model/aro_hcp/v1alpha1/break_glass_credential_resource.model
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a specific break glass credential.
+resource BreakGlassCredential {
```
vkareh, line 18: Not sure if it has been brought up, but the "break-glass" moniker no longer holds, since these are not "break glass" in the sense that they're only used as an emergency when trying to access/debug a cluster. Any chance we can rename the ARO API to be `Credential` or `AdminCredential`?

model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
vkareh: I agree with making this a `class`, but should it be named `status` instead of `status_details`? I haven't seen the use of a `status_details` struct before, so this seems to add additional disparities between resources.

model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
```diff
@@ -0,0 +1,34 @@
...
+
+// Representation of the state of a break glass credential
+struct BreakGlassCredentialState {
+    // A string value representing the credential's current state
+    Value BreakGlassCredentialStatus
```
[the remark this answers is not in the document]
vkareh: Would `BreakGlassCredentialStatusValue` or `BreakGlassCredentialStateValue` be better alternatives? That way we can use `BreakGlassCredentialStatus` for the `credentials.status` struct.

model/clusters_mgmt/v1/break_glass_credential_status_details_type.model
JameelB: +1 it would be nicer to have it as `status` but as Miguel said it would be hard to change as we will be switching it from a string to a struct in an existing api used by rosa hcp. The `status_details` was proposed as an alternative. 

The other option was using `state` as a struct but this isn't consistent with how we use this field in other resources, where it is either a string that indicates the status of the resource, or a struct that contains the `value` of the state along with other state metadata like `last_updated_timestamp`.

Accepting any alternatives here 🙏

model/aro_hcp/v1alpha1/break_glass_credential_resource.model
JameelB: We can, as long as we're in agreement of the name change. It wasn't a requirement before, but happy to change it. I don't think it should break anything as these are new endpoints 🤔

PR #1051, opened 2025-06-16
ARO-15234: Review namings of managed identities requirement types.

machi1990:
This came up as part of the review of the CS work and we noticed that we were using some generic names for the types. The naming choosen here was to make it clear that the types are about the managed identity requirement.


machi1990:
@miguelsorianod PTAL as discussed in the other thread

machi1990:
Thanks @miguelsorianod

PR #1052, opened 2025-06-23
OCM-15784: adding AWSBackupConfig type and included it in AWSShard model

cristianoveiga:


lucasponce:
LGTM

API changes are aligned with the DDR approved.

https://docs.google.com/document/d/17Wa_uTtYzuclRfPC1_xGUvCJ8FrIF_x941Hock4JesY/edit?tab=t.0#heading=h.bupciudrwmna

tzvatot (changes requested):

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -1089,6 +1089,27 @@
           }
         }
       },
+      "AWSBackupConfig": {
```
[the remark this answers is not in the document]
cristianoveiga: Ideally, it shouldn't - but we do have existing AWS types.

 Is there a way to prevent the AWS-specific types to propagate to the ARO openapi spec?
 
 cc: @ahitacat

model/clusters_mgmt/v1/aws_backup_config_type.model
```diff
@@ -0,0 +1,30 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Backup configuration for AWS clusters
+struct AWSBackupConfig {
+    // ID of the AWS Disaster Recovery (DR) account
+    AccountID           string
```
vkareh, line 20: Do we still use the `FooID` convention or have we moved to `FooId` instead? I think this should be `AccountId`, but not sure if we've adopted that entirely or if it's relegated to just variable names.

model/clusters_mgmt/v1/aws_backup_config_type.model
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
```
vkareh, line 2:
```suggestion
Copyright (c) 2025 Red Hat, Inc.
```

model/clusters_mgmt/v1/aws_backup_config_type.model
cristianoveiga: I couldn't find explicit guidelines around that. In any case, I've updated it to `AccountId`.

model/clusters_mgmt/v1/aws_backup_config_type.model
cristianoveiga: Thanks, updated to 2025.

PR #1053, opened 2025-06-24
Remove multi_az documentation for aro_hcp

mahesh-palkar:
For ARO-HCP, the multi_az flag is read-only and cannot be configured by the user. Internally, the field is unused, and as a non-pointer boolean, it defaults to false.
The reason is that once the value is mapped into the internal struct as a non-pointer, we can no longer distinguish whether it was explicitly set by the user or intended to be derived later via Azure APIs. Using a pointer would preserve this distinction but it will cause a broader impact on the codebase. Hence, using a pointer was avoided.

Ticket: https://issues.redhat.com/browse/ARO-17289
MR : https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/9973


model/clusters_mgmt/v1/cluster_type.model
```diff
@@ -70,6 +70,9 @@ class Cluster {
 	// Flag indicating if the cluster should be created with nodes in
 	// different availability zones or all the nodes in a single one
 	// randomly selected.
+	// ARO-HCP, this flag is read-only and cannot be configured by the user.
```
miguelsorianod, line 73: I would just mention:

For ARO-HCP Clusters, this attribute is unused, and the control plane is deployed in multiple availability zones when the Azure region where it is deployed supports multiple availability zones.

Internal implementation details do not need to be described at the API level documentation.-

mahesh-palkar: Thanks for the suggestion. I've updated the documentation.

vkareh (approved):
/lgtm

PR #1059, opened 2025-07-01
[ARO-17559] Add CIDR Block Access Model for Kube API-Server

nimrodshn:


openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -2058,6 +2058,13 @@
             "description": "The URL of the API server of the cluster.",
             "type": "string"
           },
+          "allowed_cidr_blocks": {
+            "description": "The CIDR blocks that are whitelisted for the API server. Currently, only supported for ARO-HCP.",
+            "type": "array",
+            "items": {
+              "type": "string"
```
miguelsorianod, line 2065: We should document the restrictions on the elements: length, format, ...

miguelsorianod: Do we foresee information about the block to have more than just the IP as the information? I am wondering if the best would be an array of string or an array of struct if we foresee this having added information in the future

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -2058,6 +2058,13 @@
             "description": "The URL of the API server of the cluster.",
             "type": "string"
           },
+          "allowed_cidr_blocks": {
+            "description": "The CIDR blocks that are whitelisted for the API server. Currently, only supported for ARO-HCP.",
```
miguelsorianod, line 2062: If this is whitelisted, does it mean that by default no access is given at all? I think this lacks definition.
Looking at the strat https://issues.redhat.com/browse/XCMSTRAT-517 it mentions:
```
    Allow a user to define allow/block lists for IP addresses or IP CIDR ranges to the API server at install time. 
    Allow updating the list of IP addresses at a later date
    Allow disabling of this feature and resort to defaults at a later date
```

Are both allow and deny lists requested? what are the requirements here?
I think whether this is the way to introduce it is not fully clear

model/clusters_mgmt/v1/cluster_api_type.model
```diff
@@ -20,4 +20,8 @@ struct ClusterAPI {
 	URL String
 	// The listening method of the API server.
 	Listening ListeningMethod
+	// The CIDR blocks that are allowed for the Cluster API server. Currently, only supported for ARO-HCP.
+	// The maximum number of CIDR blocks supported is 500. The CIDR blocks should be non-overlapping and valid.
+	// If no CIDR blocks provided all traffic is allowed to the Cluster API Server.
+	AllowedCIDRBlocks []String
```
miguelsorianod, line 26: Are we supporting IPv4, IPv6, or both? This should be specified

miguelsorianod, line 26: It would also be good to described the format. If we assume for example ipv4 it would be something like:
`<ipv4_address>/<network_mask>`

miguelsorianod, line 26: We should describe the following cases:

* What happens if I don't specify the attribute?
* Is an empty array a valid value? if not, document that. If it is, what is its behavior?

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -2058,6 +2058,13 @@
             "description": "The URL of the API server of the cluster.",
             "type": "string"
           },
+          "allowed_cidr_blocks": {
+            "description": "The CIDR blocks (An IPV4 address range in the format `<ipv4_address>/<network_mask>`) that are allowed for the Kubernetes API server. Currently, only supported for ARO-HCP based clusters.\nThe maximum number of CIDR blocks supported is 500. The CIDR blocks should be non-overlapping and valid.\nIf the field is left empty (alternatively, an empty array provided) all traffic is allowed to the Kubernetes API Server.",
```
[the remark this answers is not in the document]
nimrodshn: >Does this mean that both not setting the attribute at all and setting it with an empty array behave in the same way, which is allowing all traffic?

Yes. Correct!

>If the answer is yes, what if I want to block all traffic?

Traditionally `0.0.0.0/0` would mean "all traffic" so theoretically that would be the value to write - I would need to check if it applies in this case as well.

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -2058,6 +2058,13 @@
             "description": "The URL of the API server of the cluster.",
             "type": "string"
           },
+          "allowed_cidr_blocks": {
+            "description": "The CIDR blocks that are allowed for the Cluster API server. Currently, only supported for ARO-HCP.\nThe maximum number of CIDR blocks supported is 500. The CIDR blocks should be non-overlapping and valid.\nIf no CIDR blocks provided all traffic is allowed to the Cluster API Server.",
```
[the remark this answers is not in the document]
nimrodshn: @miguelsorianod Only K8s api server is affected by this field.

model/clusters_mgmt/v1/cluster_api_type.model
nimrodshn: @miguelsorianod  FYI - Were waiting for a response from Jerome on the matter.

openapi/aro_hcp/v1alpha1/openapi.json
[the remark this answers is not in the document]
miguelsorianod: I would interpret setting `0.0.0.0/0` in that case to mean **allow** *all* traffic, as this attribute is allowed_cidr_blocks

openapi/aro_hcp/v1alpha1/openapi.json
miguelsorianod: Understood

model/clusters_mgmt/v1/cluster_api_type.model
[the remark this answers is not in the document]
miguelsorianod: For example, the following range `10.0.0.0 – 10.255.255.255` is a private ip addresses range.
For example, the following range `11.0.0.0 – 126.255.255.255` is a public ip addresses range.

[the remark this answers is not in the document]
nimrodshn: @miguelsorianod Let me verify. We may wish to to block the entire "IANA" IP range https://en.wikipedia.org/wiki/Reserved_IP_addresses

openapi/clusters_mgmt/v1/openapi.json
```diff
@@ -15680,6 +15680,13 @@
             "description": "The URL of the API server of the cluster.",
             "type": "string"
           },
+          "allowed_cidr_blocks": {
+            "description": "The CIDR blocks (An IPV4 address range in the format `<ipv4_address>/<network_mask>`) that are allowed for the Kubernetes API server. Currently, only supported for ARO-HCP based clusters.\nThe maximum number of CIDR blocks supported is 500. The CIDR blocks should be non-overlapping and valid.\nIf the field is left empty (alternatively, an empty array provided) all traffic is allowed to the Kubernetes API Server.",
```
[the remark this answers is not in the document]
nimrodshn: >If there's no requirement to "deny all" I think that supporting both left empty and all traffic allowed might be unneded. Why not just defining one of them? Less use cases to support.

I should note that both cases allowing all traffic [e.g. left empty and `0.0.0.0/0`] are default implementation from Hypershift API so in they do not cost us anything.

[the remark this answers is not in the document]
jharrington22: We should not have a use case for deny all, if a customer wants that behavior they would instead make their kube API private (future feature).

model/clusters_mgmt/v1/cluster_api_type.model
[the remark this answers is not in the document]
nimrodshn: According to Jerome we shouldn't validate any private addressing.

openapi/aro_hcp/v1alpha1/openapi.json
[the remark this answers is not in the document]
miguelsorianod: From latest conversations in the corresponding DDR, 0.0.0.0/0 is not allowed. We should document constraints on the accepted values. This is, "0.0.0.0/0" is not allowed in this case.

model/clusters_mgmt/v1/cluster_api_type.model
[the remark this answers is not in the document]
nimrodshn: Let me verify..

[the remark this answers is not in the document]
miguelsorianod: By "Private IPs" I meant IPs that are in the private address space: https://datatracker.ietf.org/doc/html/rfc1918#section-3.

Private IP addresses are a subset of the set of reserved ip addresses.

I don't know what Jerome was referring to, and whether it included private ip addresses as part of the reserved ones. In my case I was exclusively referring to ip addresses in the private address space.

openapi/clusters_mgmt/v1/openapi.json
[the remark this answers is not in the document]
nimrodshn: Done, Ptal.

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -2474,6 +2505,10 @@
       "ClusterAPI": {
         "description": "Information about the API of a cluster.",
         "properties": {
+          "cidr_blocks_access": {
```
tzvatot, line 2508: `cidr_block_access` in singular

tzvatot (changes requested):

model/clusters_mgmt/v1/cluster_api_type.model
```diff
@@ -20,4 +20,32 @@ struct ClusterAPI {
...
+	// Otherwise, if "allow_all" is provided than 'values' list should be omitted.
+	Mode CIDRBlocksAccessMode
+	// The 'values' list should contain a CIDR blocks list (An IPV4 address range in the format `<ipv4_address>/<network_mask>`). 
+	// The maximum number of CIDR blocks supported is 500. The CIDR blocks should be non-overlapping and valid.
+	// The value "0.0.0.0/0" is not considered a valid value, as the user can use "allow_all" mode to indicate this behavior.
+	Values []String
+}
+
+// CIDR Blocks Access modes.
+enum CIDRBlocksAccessMode {
```
tzvatot, line 45: @miguelsorianod mentioned using enums will break the client if we introduce another value.

tzvatot (changes requested):

model/clusters_mgmt/v1/cluster_api_type.model
```diff
@@ -20,4 +20,24 @@ struct ClusterAPI {
 	URL String
 	// The listening method of the API server.
 	Listening ListeningMethod
+	
+	// Describes the access policy for CIDR Blocks to the Kubernetes API server
+	CIDRBlocksAccess CIDRBlocksAccess
```
tzvatot, line 25: I noticed we have a "network" struct on the cluster, that have some related cidr settings. I assume the choice to put it under "api" is because it's relevant ONLY for kube-api-server access. @miguelsorianod @nimrodshn is that correct?

tzvatot:
assuming the respond to https://github.com/openshift-online/ocm-api-model/pull/1059#discussion_r2260023276 is yes, LGTM

model/clusters_mgmt/v1/cluster_api_type.model
nimrodshn: Yes, correct.

openapi/clusters_mgmt/v1/openapi.json
```diff
@@ -13657,6 +13657,29 @@
           }
         }
       },
+      "CIDRBlocksAccess": {
```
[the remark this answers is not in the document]
tzvatot: Should be singular, like "nail gun" (for nails) and "screw driver" (for screws) and so "cidr block access" for cidr blocks.

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -1541,6 +1541,29 @@
           }
         }
       },
+      "CIDRBlockAccess": {
+        "description": "Describes the CIDR Block access policy to the Kubernetes API server.\nCurrently, only supported for ARO-HCP based clusters.",
```
miguelsorianod, line 1545: We need to describe that when not provided all access is allowed

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -1541,6 +1541,29 @@
...
+        "properties": {
+          "allow": {
+            "$ref": "#/components/schemas/CIDRBlockAllowAccess"
+          }
+        }
+      },
+      "CIDRBlockAllowAccess": {
+        "properties": {
+          "mode": {
+            "description": "There are two modes: \"allow_all\" and \"allow_list\"; if \"allow_list\" is provided than a non-empty 'values' list must be provided.\nOtherwise, if \"allow_all\" is provided than 'values' list should be omitted.",
```
miguelsorianod, line 1555: than -> then

PR #1069, opened 2025-07-14
OCM-15786 | feat: adding methods to enable/disable backups of control planes

cristianoveiga:


tzvatot (changes requested):

model/clusters_mgmt/v1/backup_type.model
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of a Backup.
+struct Backup {
+	// Indicates whether the backup is enabled
+	Enabled Boolean
```
tzvatot, line 20: Please note the API coding guidelines: https://gitlab.cee.redhat.com/service/api-guidelines#avoid-boolean-fields-and-binary-state-wording

tzvatot (changes requested):

model/clusters_mgmt/v1/backup_type.model
cristianoveiga: Added the BackupStateValue type.

model/clusters_mgmt/v1/backup_type.model
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Representation of a Backup.
+struct Backup {
+	// Indicates the state of the backup
+	BackupState BackupStateValue
```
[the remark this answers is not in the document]
cristianoveiga: Done.

model/clusters_mgmt/v1/backup_state_value_type.model
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// BackupStateValue represents the state of a backup
+enum BackupStateValue {
```
tzvatot, line 18: @miguelsorianod had an interesting point when we last discussed about the enryption at host feature (enabled/disabled): if we use enum, and we later expand it, it's considered an API breaking change on the client side. However, if we use string instead (although they are represented the same way in the json as enum), it's not considered a breaking change and will not necessary require client to be modified. 

Hence, I suggest to change it to be string and document the allowed values.

cristianoveiga: Done.

tzvatot (changes requested):

model/clusters_mgmt/v1/backup_state_value_type.model
cristianoveiga: good catch, thanks!

 I forgot to delete the model file.

tzvatot (approved):
/lgtm

vkareh (approved):
/lgtm

PR #1077, opened 2025-07-23
Add tag make targets

nimrodshn:
- Adds a new `release` make target 
- Adds some instructions about releasing a new version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,48 @@ clean:
         clientapi/webrca \
 		$(NULL)
 
+.PHONY: validate-version
+validate-version:
+ifndef VERSION
+	$(error VERSION is required. Usage: make tag VERSION=v0.0.426)
+endif
+	@# Validate that we're on the main branch
+	@CURRENT_BRANCH=$$(git rev-parse --abbrev-ref HEAD); \
+	if [ "$$CURRENT_BRANCH" != "main" ]; then \
+		echo "Error: Tags must be created from the main branch. Currently on $$CURRENT_BRANCH"; \
+		exit 1; \
+	fi
+	@# Validate version format (vX.Y.Z)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -108,6 +108,26 @@ clean:
         clientapi/webrca \
 		$(NULL)
 
+.PHONY: tag
+tag:
```
tzvatot, line 112: How do we validate that the tag is created from the `main` branch?

tzvatot: [[ "$(git symbolic-ref --short HEAD 2>/dev/null)" == "main" ]] && echo "On main branch" || echo "Not on main branch"

Makefile
```diff
@@ -108,6 +108,48 @@ clean:
...
+
+.PHONY: tag_and_push_modules
+tag_and_push_modules: validate-version
+	@for module in $(MODULES); do \
+		git tag -a -m 'Release $(VERSION)' $$module/$(VERSION); \
+		git push upstream $$module/$(VERSION); \
+	done 
+
+.PHONY: release
+release: tag_and_push_modules validate-version
```
tzvatot, line 149: Since `tag_and_push_modules` depends on `validate-version`  do we need both dependencies here?

Makefile
```diff
@@ -108,6 +108,48 @@ clean:
         clientapi/webrca \
 		$(NULL)
 
+MODULES = model clientapi metamodel_generator 
+.PHONY: validate-version
+validate-version:
```
tzvatot, line 113: This target is written with dash ("-") while the others are using lower dash '_'. Can we align for consistency? It seems convention is dash.

tzvatot (changes requested):

tzvatot (approved):
/lgtm

PR #1078, opened 2025-07-28
[DRAFT] OCM-16333: add SkipValidations filed into class ControlPlaneUpgradePolicy

marcolan018:
related DDR requests this api change: https://docs.google.com/document/d/1DWNR9bvDb41Cdt2TX6E8TNmz-KSy9nRariujKSBHznU/edit?tab=t.0

for the epic https://issues.redhat.com/browse/OCM-16325 to provide API support HCP force upgrade, we propose to add `skipValidations` field into `ControlPlanUpgradePolicy` so that users can specify the validations which should be skipped before upgrade

@robpblake @etabak PTAL, thanks

### DIFF
--- a/model/clusters_mgmt/v1/control_plane_upgrade_policy_type.model
+++ b/model/clusters_mgmt/v1/control_plane_upgrade_policy_type.model
@@ -45,4 +45,7 @@ class ControlPlaneUpgradePolicy {
 
 	// State of the upgrade policy for the hosted control plane.
 	State UpgradePolicyState
+
+	// Validations skipped before upgrade
+	SkipValidations []String
 }
--- a/clientapi/clustersmgmt/v1/control_plane_upgrade_policy_builder.go
+++ b/clientapi/clustersmgmt/v1/control_plane_upgrade_policy_builder.go
@@ -34,6 +34,7 @@ type ControlPlaneUpgradePolicyBuilder struct {
 	nextRun                    time.Time
 	schedule                   string
 	scheduleType               ScheduleType
+	skipValidations            []string
 	state                      *UpgradePolicyStateBuilder
 	upgradeType                UpgradeType
 	version                    string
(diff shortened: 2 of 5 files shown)


model/clusters_mgmt/v1/control_plane_upgrade_policy_type.model
```diff
@@ -45,4 +45,7 @@ class ControlPlaneUpgradePolicy {
 
 	// State of the upgrade policy for the hosted control plane.
 	State UpgradePolicyState
+
+	// Validations skipped before upgrade
+	SkipValidations []String
```
tzvatot, line 50: How will a customer know which validations are available to be skipped? is there a dedicated endpoint to expose it?

tzvatot (changes requested):

model/clusters_mgmt/v1/control_plane_upgrade_policy_type.model
tzvatot: As you know the API have a tendency to stay for a very long period of time.

The values (or "names) of the validations, may change over time, as validations are getting added/removed/modified.
Which means, that if SRE set some skip validations to a cluster, we should keep in mind that by the time the cluster actually goes to upgrade, those validations may be stale - their name, or functionality, may have changed.
In addition, one cluster's validations may differ than another - which may cause confusion. Imagine a cluster with a certain configuration, that has a certain skip validation, and SRE copy that skip validation to another cluster that has a different configuration (say from HCP cluster to non-HCP) to try to force upgrade - failing to do so. 
Without knowing in advance which are the allowed validation per cluster, it may be very difficult for anyone to try and guess which are the suitable validations. And even if they get it right, the code changes and it may become stale the next time they will try it.

What if we use the `--force` flag approach that we have for destroying a cluster, replicating it to upgrades? 
Would SRE really care about which validations they actually want to skip? why hand-pick them to begin with? 

Another suggestion:
When upgrade policy API was design we discussed the option of creating another type of upgrade which is CVE, that SRE will push. What if introduce another type of upgrade, that can only be applied by SRE, which will indicate skipping any validations?

marcolan018:
close this PR
due to the latest design, no API change is required

PR #1082, opened 2025-08-06
Remove old field osDiskSseEncryptionSetResourceId in AzureNodePool model

anshulvermapatel:
The field `osDiskSseEncryptionSetResourceId ` has been shifted to a nested struct `AzureNodePoolOsDisk`. This is to remove the old remaining field in the AzureNodePool struct.
It can be removed as it is an attribute specific to ARO-HCP and the attribute was not used anywhere yet.


miguelsorianod:
We can remove this because it is an attribute specific to ARO-HCP and the attribute was not used anywhere yet.

PR #1083, opened 2025-08-06
Added changes from the last tag till main in CHANGES.md

anshulvermapatel:
Updated CHANGED.md with the latest change done since the recent tag.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,16 +2,22 @@
 
 This document describes the relevant changes between releases of the API model.
 
- ## 0.0.426 Jul 23 2025
- - no changes
+## 0.0.427 Aug 06 2025
+- Remove `osDiskSseEncryptionSetResourceId` field from type `AzureNodePoolOsDisk`
+- Add methods to enable/disable backups of control planes
+- Fix aws capacity reservation class to struct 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -2,6 +2,12 @@
 
 This document describes the relevant changes between releases of the API model.
 
+## 0.0.427 Aug 06 2025
```
tzvatot, line 5: While updating the changes - it would also make sense to fill up the gap for the other releases that we've missed.

tzvatot (changes requested):

CHANGES.md
miguelsorianod: Can that be addressed as a follow-up MR? We'd like to move forward with having a new release with other changes

miguelsorianod: It is more we would like to avoid investing the time on gathering the conceptual changes right now

tzvatot: AI for the rescue? https://github.com/openshift-online/ocm-api-model/pull/1086

tzvatot (changes requested):

tzvatot (approved):
/lgtm

PR #1084, opened 2025-08-06
OCM-16333 | feat: add ControlPlaneCVE to UpgradeType

marcolan018:
This MR addes a new type `ControlPlaneCVE` to UpgradeType enum.

```
enum UpgradeType {
        ....
        ...
	@json(name = "ControlPlaneCVE")
	ControlPlaneCVE
}
```
related DDR: https://docs.google.com/document/d/1DWNR9bvDb41Cdt2TX6E8TNmz-KSy9nRariujKSBHznU/edit?tab=t.0

### DIFF
--- a/clientapi/clustersmgmt/v1/upgrade_type_type.go
+++ b/clientapi/clustersmgmt/v1/upgrade_type_type.go
@@ -29,6 +29,8 @@ const (
 	UpgradeTypeAddOn UpgradeType = "ADDON"
 	// Control plane upgrade, relevant only for hosted control plane clusters.
 	UpgradeTypeControlPlane UpgradeType = "ControlPlane"
+	// An upgrade required for security reasons.
+	UpgradeTypeControlPlaneCVE UpgradeType = "ControlPlaneCVE"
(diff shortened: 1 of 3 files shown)


clientapi/clustersmgmt/v1/upgrade_type_type.go
```diff
@@ -23,6 +23,8 @@ package v1 // github.com/openshift-online/ocm-api-model/clientapi/clustersmgmt/v
 type UpgradeType string
 
 const (
+	// Control plane upgrade without validations, relevant only for hosted control plane clusters.
+	UpgradeTypeCVE UpgradeType = "CVE"
```
tzvatot, line 27: Since we already have control_plane type, and currently this CVE type is only for control_plane, I think we better differentiate so not to create confusion with what is actually upgrading. Suggesting "ControlPlaneCVE".

tzvatot (changes requested):

tzvatot:
@nimrodshn do we need to explicitly add it to ARO HCP too?

PR #1093, opened 2025-08-27
OCM-17726 | Add resource-scoped permissions for WIF configurations

ckandag:
https://issues.redhat.com/browse/OCM-17726

Per DDR attached to the issue above,  this PR makes model changes to include resource_bindings
for scoped permissions.

- Add ResourceBindings field to WifRole struct
- Create new WifResourceBinding struct with Type and Name fields


ckandag:
@vkareh  @ciaranRoche  would you please take a look

PR #1098, opened 2025-09-04
OCM-689: Added AccountGroup and AccountGroupAssignment APIs for RBAC

Alcamech:


model/accounts_mgmt/v1/account_group_assignment_resource.model
```diff
@@ -0,0 +1,33 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a specific account group assignment.
+resource AccountGroupAssignment {
```
tirthct, line 18: Do we need a List here?

tirthct: Ah, this is a singular assignment endpoint. Not needed!

model/accounts_mgmt/v1/account_group_assignment_managed_by_type.model
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright (c) 2019 Red Hat, Inc.
```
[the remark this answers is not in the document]
Alcamech: done

model/accounts_mgmt/v1/role_binding_type.model
```diff
@@ -20,12 +20,14 @@ class RoleBinding {
 	SubscriptionID String
 	Account Account
 	AccountID String
+  AccountGroup AccountGroup
+  AccountGroupID String
```
cristianoveiga, line 24: nit: this seems to be misaligned. Maybe tabs vs spaces ?

model/accounts_mgmt/v1/role_binding_type.model
```diff
@@ -20,12 +20,14 @@ class RoleBinding {
...
+	AccountGroupID String
 	Organization Organization
 	OrganizationID String
 	Role Role
 	RoleID String
 	ConfigManaged Boolean
 	CreatedAt Date
 	UpdatedAt Date
-	ManagedBy String
+	ManagedBy RoleBindingManagedBy
```
cristianoveiga, line 32: @Alcamech I wonder if it would be safer to leave this as string in order to preserve backwards compatibility.

@vkareh any thoughts on that ?

tzvatot: > Tricky - since enums are strings, it's technically not a breaking change in the API itself. However, it will break any client that bumps the SDK version. Do we have any client using the SDK on this field? I know I added support for `rosa` _years_ ago, but I don't think I ever merged it.
> 
> @tzvatot any thoughts? I think I'm okay with this change, since the worst case scenario is that a client bumps the SDK, which will break the build, forcing an update to the new API. Annoying (assuming anyone uses this) but it won't break any existing build, which is the main concern.

The main concern is you pointed out is breaking the client. If rosa cli use it, it may be harder to upgrade. It boils down the to question of - how hard would it be to upgrade the client, from a customer perspective. We don't know. In some deployments, customers may be very limited in how or when they upgrade (fedramp?). I would suggest leave it as a string. The cost of making a mistake, does not worth the value.

Alcamech: Reverted back to a string

PR #1157, opened 2026-04-29
OCM-23443 | feat: add zero_egress field to aws spec

marcolan018:
https://redhat.atlassian.net/browse/OCM-23443

this MR provides a new field `zero_egress` into aws spec. This will provide the zero egress related info of the cluster.
Currently it includes the `no_proxy_default_domains`. A sample is like below
```
{       
    "kind":"Cluster",
    "id":"...",
    "name":"...",     
     ...
     ...                                                                                                                                                                                                     
    "aws": {
          "account_id": "...",
          "subnet_ids": [...],
          "zero_egress": {
            "enabled": true,
            "no_proxy_default_domains": [
              "s3.dualstack.{region}.amazonaws.com",                                                                                                                                                                                                            
```

### DIFF
--- a/clientapi/clustersmgmt/v1/aws_builder.go
+++ b/clientapi/clustersmgmt/v1/aws_builder.go
@@ -43,13 +43,14 @@ type AWSBuilder struct {
 	subnetIDs                              []string
 	tags                                   map[string]string
 	vpcEndpointRoleArn                     string
+	zeroEgress                             *ZeroEgressBuilder
 	privateLink                            bool
 }
 
 // NewAWS creates a new builder of 'AWS' objects.
 func NewAWS() *AWSBuilder {
 	return &AWSBuilder{
-		fieldSet_: make([]bool, 22),
+		fieldSet_: make([]bool, 23),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


lucasponce:
I don't think "service_inquiries" makes sense, as it's a "verb" and we're "hiding" a resource in the payload.

Why don't simply add the ZeroEgress as part of the AWS payload?

https://github.com/openshift-online/ocm-api-model/blob/main/model/clusters_mgmt/v1/aws_type.model

Or if you want to make it more robust, you can add a ZeroEgress struct with an Enabled/Disabled and inside the details that should be populated if enabled.

The API model has multiple examples with that pattern.

marcolan018:
@lucasponce thanks for the comments
can you help to take another look on the latest changes? Thanks

lucasponce:
LGTM

The only comment is that we need to ensure that the payload 

```
"zero_egress": {
     "enabled": false
}
```

Should be semantically compatible with the absence of the "zero_egress" to not break any client.

But +1 on this approach, given the context with the direction and challenges we have to take in the API it looks safer to me.

gdbranco (approved):

marcolan018:
@lucasponce 
  The API model definition already provided the compatibility
  1. The ZeroEgress field in AWS is optional (returns nil when absent)                                                                                                                                               
  2. The Enabled() method returns false when the field is unset (even if the object exists)                                                                                                                        
  3. Services must use nil-safe checks when accessing the field

PR #1164, opened 2026-05-13
add AzureContainerRegistryPullCredentials type to ARO HCP API model

twolff-gh:
## Description

Required for the ACR Pull with MI feature to be plumbed.

Data flow:

1. RP receives acrPullIdentity (ARM resource ID of a user-assigned managed identity)
2. RP converts to Azure.ImageRegistryCredentials.ManagedIdentityResourceId via OCM SDK
3. CS stores it and sets it on the HyperShift HostedCluster CR
4. HyperShift attaches the MI to worker VMSS and writes it into worker cloud.conf
5. Kubelet's ACR credential provider uses the MI to authenticate image pulls from ACR

Why a new type: No existing OCM model type carries a managed identity resource ID for image registry authentication.

Feature https://redhat.atlassian.net/browse/ARO-24037
Implementation https://redhat.atlassian.net/browse/ARO-25891

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or tooling change

## Verification

- [x] `make check` passes (model syntax validation)
- [x] `make lint` passes (indentation enforcement)

## Checklist

- [ ] Model files follow the project's DSL conventions (see [README](../README.md#concepts))
- [ ] Documentation comments use Markdown format
- [ ] Generated `clientapi/` and `openapi/` are updated if model changed (`make update`)

### DIFF
--- a/model/aro_hcp/v1alpha1/azure_container_registry_credentials_type.model
+++ b/model/aro_hcp/v1alpha1/azure_container_registry_credentials_type.model
@@ -1,0 +1,29 @@
+/*
+Copyright (c) 2026 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Azure Container Registry credentials configuration for an ARO HCP Cluster.
+// Configures authentication for container image pulls from Azure Container
+// Registry (ACR) on Data Plane worker nodes.
+struct AzureContainerRegistryCredentials {
+	// The credential type used for ACR image pulls on Data Plane worker nodes.
+	// Required.
+	Type AzureContainerRegistryCredentialType
+
+	// The user-assigned managed identity used for ACR image pulls
+	// on Data Plane worker nodes.
+	// Required when type is ManagedIdentity.
+	ManagedIdentity AzureUserAssignedManagedIdentity
+}
--- a/model/aro_hcp/v1alpha1/azure_type.model
+++ b/model/aro_hcp/v1alpha1/azure_type.model
@@ -118,6 +118,31 @@ struct Azure {
 	// Readonly
 	OidcIssuerUrl String
 
+	// Configures authentication for container image pulls from Azure
+	// Container Registry (ACR) on Data Plane worker nodes. Only ACR
+	// registries (*.azurecr.io, *.azurecr.cn, *.azurecr.de, *.azurecr.us) are
+	// supported; non-ACR registries are not affected by this
+	// configuration. The identity is not bound to a specific ACR
+	// instance — it authenticates to any ACR the identity has the
+	// AcrPull role assigned on. The customer controls which ACR
+	// instances the identity can pull from via Azure RBAC.
+	// When set, the managed identity is attached to worker virtual
+	// machines and its resource ID is written into the worker cloud
+	// provider config so kubelet's ACR credential provider can
+	// authenticate without image pull secrets.
+	// When not set, no managed identity is attached to worker virtual
+	// machines for ACR and no identity is written into the worker
+	// cloud provider config. The ACR credential provider on worker
+	// nodes has no identity to authenticate with, so image pulls from
+	// private ACR registries will fail unless the workload provides
+	// image pull secrets or another authentication mechanism. Public
+	// ACR images (anonymous pull enabled) are not affected.
+	// Changing this value will trigger a rollout for all existing
+	// NodePools in the cluster.
+	// Optional.
+	// Mutable.
+	ContainerRegistry AzureContainerRegistry
+
     // Vnet Integration Subnet Resource ID is the Azure Resource ID of a pre-existing Azure
     // Subnet used for SWIFT networking (Azure Container Networking Interface).
     // This subnet is delegated to Microsoft.RedHatOpenShift/hcpOpenShiftClusters
--- a/openapi/aro_hcp/v1alpha1/openapi.json
+++ b/openapi/aro_hcp/v1alpha1/openapi.json
@@ -4026,6 +4026,10 @@
       "Azure": {
         "description": "Microsoft Azure settings of a cluster.",
         "properties": {
+          "container_registry": {
+            "description": "Configures authentication for container image pulls from Azure\nContainer Registry (ACR) on Data Plane worker nodes. Only ACR\nregistries (*.azurecr.io, *.azurecr.cn, *.azurecr.de, *.azurecr.us) are\nsupported; non-ACR registries are not affected by this\nconfiguration. The identity is not bound to a specific ACR\ninstance — it authenticates to any ACR the identity has the\nAcrPull role assigned on. The customer controls which ACR\ninstances the identity can pull from via Azure RBAC.\nWhen set, the managed identity is attached to worker virtual\nmachines and its resource ID is written into the worker cloud\nprovider config so kubelet's ACR credential provider can\nauthenticate without image pull secrets.\nWhen not set, no managed identity is attached to worker virtual\nmachines for ACR and no identity is written into the worker\ncloud provider config. The ACR credential provider on worker\nnodes has no identity to authenticate with, so image pulls from\nprivate ACR registries will fail unless the workload provides\nimage pull secrets or another authentication mechanism. Public\nACR images (anonymous pull enabled) are not affected.\nChanging this value will trigger a rollout for all existing\nNodePools in the cluster.\nOptional.\nMutable.",
+            "$ref": "#/components/schemas/AzureContainerRegistry"
+          },
           "etcd_encryption": {
             "description": "Etcd encryption configuration.\nIf not specified, etcd data is encrypted with platform managed keys.\nCurrently etcd data encryption is only supported with customer managed keys.\nCreating a cluster with platform managed keys will result in a failure creating the cluster.",
             "$ref": "#/components/schemas/AzureEtcdEncryption"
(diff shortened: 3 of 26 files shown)


model/aro_hcp/v1alpha1/azure_type.model
```diff
@@ -118,6 +118,9 @@ struct Azure {
 	// Readonly
 	OidcIssuerUrl String
 
+	// Azure image registry credentials configuration for ACR image pulling.
```
zgalor: The field comment lacks `// Optional` or `// Required during creation`  `// Immutable` annotations that other fields in this file use (see lines 20-22, 73-74, 96-97, etc.)

model/aro_hcp/v1alpha1/azure_image_registry_credentials_type.model
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+// Azure image registry credentials configuration for an ARO HCP Cluster.
+// Configures a user-assigned managed identity for authenticating container
+// image pulls from Azure Container Registry (ACR).
+struct AzureImageRegistryCredentials {
+    // The Azure Resource ID of the user-assigned managed identity that has
+    // the permissions required for ACR image pulling.
+    ManagedIdentityResourceId String
```
zgalor: The ManagedIdentityResourceId field lacks `// Required` `// Immutable` `// Readonly` annotations

zgalor (changes requested):
Overall lgtm. some field annotations missing

zgalor (approved):

model/aro_hcp/v1alpha1/azure_type.model
```diff
@@ -118,6 +118,10 @@ struct Azure {
...
+	// Optional.
+	ImageRegistryCredentials AzureImageRegistryCredentials
```
machi1990:
```suggestion
	AzureContainerRegistryPullCredentials AzureContainerRegistryPullCredentials
```

model/aro_hcp/v1alpha1/azure_container_registry_credentials_type.model
machi1990, line 1: model/aro_hcp/v1alpha1/azure_container_registry_pull_credentials_type.model

model/aro_hcp/v1alpha1/azure_image_registry_credentials_type.model
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
```
machi1990:
```suggestion
Copyright (c) 2026 Red Hat, Inc.
```

model/aro_hcp/v1alpha1/azure_image_registry_credentials_type.model
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Azure image registry credentials configuration for an ARO HCP Cluster.
+// Configures a user-assigned managed identity for authenticating container
+// image pulls from Azure Container Registry (ACR).
+struct AzureImageRegistryCredentials {
```
machi1990:
```suggestion
struct AzureContainerRegistryPullCredentials {
```

machi1990 (changes requested):

miguelsorianod:
@twolff-gh @zgalor @machi1990 do we have context on whether this belongs to a particular data plane operator in the cluster?

For operators based identities we should not have the identity information in its own section, but in the operators authentication one.

Aside from that, do we have a clear idea on what built-in roles are going to be used? are we introducing permissions validation around it?

Is this a new *required* identity or is it optional? if it's required, what's our strategy for our existing clusters?

machi1990:
> @twolff-gh @zgalor @machi1990 do we have context on whether this belongs to a particular data plane operator in the cluster?

It is a data plane operator but I don't believe that this is used by any operator to perform its action. It's going to be wired onto a cloud.conf file and attached to each VM when node pools are created. This way the identity seamless can be used to pull images from ACRs once it has PullPermission (something a user has to set up).

> Aside from that, do we have a clear idea on what built-in roles are going to be used? are we introducing permissions validation around it?

The ACRPull permission is needed.
In the CS MR, I left comment around the inflights that are needed.

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -4076,6 +4080,35 @@
           }
         }
       },
+      "AzureContainerRegistry": {
+        "description": "Azure Container Registry configuration for an ARO HCP Cluster.",
```
miguelsorianod: We should improve the documentation about this:

* To what ACR registry is the pull going to happen
* Is it pulls from the control plane side or from the data plane side? (worker nodes)
* What happens when not set?

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -4076,6 +4080,35 @@
...
+        "description": "The type of credential used for Azure Container Registry image pulls.",
+        "type": "string",
+        "enum": [
+          "ManagedIdentity"
+        ]
+      },
+      "AzureContainerRegistryCredentials": {
+        "description": "Azure Container Registry credentials configuration for an ARO HCP Cluster.\nConfigures authentication for container image pulls from Azure Container\nRegistry (ACR).",
+        "properties": {
+          "managed_identity": {
```
miguelsorianod, line 4102: Is it immutable? document appropriately if so

miguelsorianod, line 4102: Document that its resourcegroup cannot be the same as the one specified for managed resource group (see azure_operators_authentication_types.model for example doc)

miguelsorianod, line 4102: Document that it must belong to the same subscription as the cluster

miguelsorianod, line 4102: Document that it must be in the same location as the cluster

miguelsorianod, line 4102: We should implement all the validations for all the restrictions that we impose and we document

openapi/aro_hcp/v1alpha1/openapi.json
miguelsorianod: The new doc changes have improved the understanding, but it's still not clear to me:
* To what ACR registry is the pull going to happen
* What happens when not set? it's now described as "When not set, worker nodes use the default Azure image pull mechanism without a user-assigned managed identity for ACR authentication." . What is "default Azure image pull mechanism without a user-assigned managed identity for ACR authentication" ?

twolff-gh: Sounds good :+1: 
Thank you :pray: 
I hope it is clearer now. Let know if you need me to further clarify anything.

openapi/aro_hcp/v1alpha1/openapi.json
twolff-gh: Added  "Changing this value will trigger a rollout for all existing NodePools." and adding // Mutable.

twolff-gh: Do we add validations here or in CS and the RP?

twolff-gh: Or both?

twolff-gh: Adding. Would a customer MI placed there be destroyed with the cluster?

twolff-gh: Adding for consistency with the operator MI pattern.

Technically the MI Entra ID service principal is tenant scoped, so cross-sub works for ACR auth. Does constraining to the same subscription simplify RP validation (existence check, assign/action RBAC).
Would we ever relax this later?

And Hypershift doesn't need this information does it? It should be able to attach whatever it provided? (Passing validations further up)

twolff-gh: Again here, more clarification would be great :pray: 

The MI's service principal is global in Entra ID so location doesn't technically matter for ACR token acquisition, but does matching the clusters location aligns with the existing platform pattern?

twolff-gh: @bennerv cc

twolff-gh: @machi1990 Would this be part of some of the additional validation you were suggesting?

twolff-gh: Or are you talking about validations at the API level here?
(sorry for the follow-on comments :pray: )

miguelsorianod: Validations at both the RP and CS level

miguelsorianod: > The MI's service principal is global in Entra ID so location doesn't technically matter for ACR token acquisition, but does matching the clusters location aligns with the existing platform pattern?

The service principal is global but managed identities are regional. ARO-HCP is a regional service so we should document and ensure that it must belong to the same location as the cluster

miguelsorianod: The service principal is tenant scoped but the managed identity is a resource that lives in a location, within a subscription, within a resourcegroup

miguelsorianod: The customer can't really create the MI in the MRG as they don't have permissions to do so. They can however specify a resource id string that contains the MRG.

miguelsorianod: If it's mutable, how do we determine that the change has occurred and completed successfully?

openapi/aro_hcp/v1alpha1/openapi.json
miguelsorianod: If enabling/disabling of the functionality is mutable, how do we determine that the change has occurred and completed successfully?

machi1990: @twolff-gh in relation to 
> To what ACR registry is the pull going to happen

It might be good to document in this API and the ARM one that the list of ACRs are  `*.` to all possible azures acrs - see https://github.com/openshift/machine-config-operator/blob/ee994cd50b06fbe93bcf20e769c1fef405a10a90/templates/common/azure/files/etc-kubernetes-credential-providers-acr-credential-provider.yaml#L11

miguelsorianod: Hi,

Thanks for the changes.

It's still not clear to me:
* To what ACR registry is the pull going to happen
* What happens when not set? it's now described as "When not set, worker nodes use the default Azure image pull mechanism without a user-assigned managed identity for ACR authentication." . What is "default Azure image pull mechanism without a user-assigned managed identity for ACR authentication" ?

miguelsorianod: Are the credentials going to be valid only for ACR registries, or also to other registries that those might have access to?

twolff-gh: Since the MIs arent bound to a specific ACR instance, the acr-credential-provider binary on worker nodes authenticates to any ACR whose FQDN matches *.azurecr.io (or *.azurecr.cn / *.azurecr.us).

Sequence: kubelet need to pull an image from a matching registry, the credential provider calls imds using the configured MI, gets a token, exchanges it for an ACR refresh token.

Customer controls ACR instances the MI can pull via AcrPull role on each target ACR. No registry URL configured in the cluster, only identity + RBAC scoped.

When ContainerRegistry isnt set:
- No UAMI attached to worked VMSS
- No ID written to cloud.conf
- acr-credential-provider binary on node has no ID to call imds with
- Image pulls form private ACR will fail with error / public images not affected

I can clarify all this. Let me know if you have any mor comments on the above :pray: 

Also,
From the Hypershift side, it only works with ACR registries. The possible registries are scoped to only Azure. By azure design, theyre scoped to tenant only. The design is essentially CS + Hypershift enables and constrains to azurecr only and allows customer to determine which MIs to use and configure their own rbac.

openapi/aro_hcp/v1alpha1/openapi.json
twolff-gh: Should we be able to detect through nodepool status conditions? 

https://github.com/openshift/hypershift/pull/8472/changes#diff-11b7c56a5faf0c5e2154fc35474456e2aac1b6ea8e22e5bbda6c3a3996d91afaR478

Triggering a rollout should make the behavior consistent. Lmk

openapi/aro_hcp/v1alpha1/openapi.json
twolff-gh: I updated the comment doc based on the above comment. Lmk if it is clearer

openapi/aro_hcp/v1alpha1/openapi.json
twolff-gh: Made a quick ticket https://redhat.atlassian.net/browse/ARO-27684
Anything else to add to the doc or the model here for validation or otherwise?

miguelsorianod: Everything that has been documented as API contract should be validated

openapi/aro_hcp/v1alpha1/openapi.json
miguelsorianod: Hi Todd,

Yes that helps understand the functionality.

miguelsorianod: After talking to @machi1990 , it seems it's still not fully clear what are the restrictions around mutability and we found that disabling in day2 is not being handled for example. He's made me aware you will follow-up to make sure we understand the behavior and the intended behavior, fixing any issues, if there are.

If there are changes we should update the doc (and validations) in both RP and CS API as a follow-up MR.

openapi/aro_hcp/v1alpha1/openapi.json
bennerv: > Document that it must belong to the same subscription as the cluster

It doesn't have to be.  Just has to be in the same tenant as the cluster, otherwise it won't attach and work properly.

bennerv: > Document that it must be in the same location as the cluster

Not a real restriction.  Do we validate this for control plane and data plane identities today?  @twolff-gh we probably should have an e2e test for this when it's implemented.

PR #1173, opened 2026-06-15
fix: relax subscription and location constraints on ACR pull managed identity

twolff-gh:
The managed identity for ACR pulls is customer-owned and passed through to CAPZ. Azure enforces regional isolation via the MI's isolationScope property, and cross-subscription attachment is supported within the same tenant. Remove same-subscription and same-location constraints from the model, keeping only same-tenant and different-MRG requirements.

Feature https://redhat.atlassian.net/browse/ARO-24037
Implementation https://redhat.atlassian.net/browse/ARO-25891

## Type of Change

- [X] Documentation update

## Verification

- [X] `make check` passes (model syntax validation)
- [X] `make verify` passes (generated code matches model)
- [X] `make lint` passes (indentation enforcement)

## Checklist

- [X] Model files follow the project's DSL conventions (see [README](../README.md#concepts))
- [X] Documentation comments use Markdown format
- [X] Generated `clientapi/` and `openapi/` are updated if model changed (`make update`)

### DIFF
--- a/model/aro_hcp/v1alpha1/azure_user_assigned_managed_identity_type.model
+++ b/model/aro_hcp/v1alpha1/azure_user_assigned_managed_identity_type.model
@@ -19,15 +19,6 @@ struct AzureUserAssignedManagedIdentity {
 	// The Azure Resource ID of the Azure User-Assigned Managed Identity.
 	// The managed identity represented must exist before creating or
 	// updating the cluster.
-	// The Azure Resource Group Name specified as part of the Resource ID
-	// must belong to the Azure Subscription specified in `.azure.subscription_id`,
-	// and in the same Azure location as the cluster's region.
-	// The Azure Resource Group Name specified as part of the Resource ID
-	// must be a different Resource Group Name than the one specified in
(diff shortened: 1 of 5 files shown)


model/aro_hcp/v1alpha1/azure_user_assigned_managed_identity_type.model
miguelsorianod, line 18: An issue with this is that we've defined the struct as a general `AzureUserAssignedManagedIdentity`. If there are other places where we use MIs in the future and the developers define the attributes as using this type (as it has a broad/general naming) then they inherit the description and restrictions of it, which might not necessarily be the case. We should avoid that.

miguelsorianod: I think what we can do instead, if we really want to have a general `AzureUserAssignedManagedIdentity` type that people can reuse is: do not document any restrictions that might change between uses of it. Document the restrictions on the documentation of the attribute that uses the type

PR #1175, opened 2026-06-23
ARO-HCP: Add Resources subresource to Cluster resource

nimrodshn:
## Description

JIRA: https://redhat.atlassian.net/browse/ARO-27920

## Type of Change

- [x] Documentation update
- [ ] CI/CD or tooling change

## Verification

- [ ] `make check` passes (model syntax validation)
- [x] `make lint` passes (indentation enforcement)

## Checklist

- [ ] Model files follow the project's DSL conventions (see [README](../README.md#concepts))
- [ ] Documentation comments use Markdown format
- [ ] Generated `clientapi/` and `openapi/` are updated if model changed (`make update`)

## New Features
* Added a new API endpoint to retrieve a list of cluster resources currently in an error state
* New Resources subresource now available, providing explicit management capabilities for cluster resources

### DIFF
--- a/model/aro_hcp/v1alpha1/resources_resource.model
+++ b/model/aro_hcp/v1alpha1/resources_resource.model
@@ -1,0 +1,24 @@
+/*
+Copyright (c) 2026 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a collection of resources for a cluster
+resource Resources {
+    // Retrieves a list of resources for a cluster in error state
+    method Get {
+        // List of cluster resources
+        out Body ClusterResources
+    }
+}
--- a/model/aro_hcp/v1alpha1/cluster_resource.model
+++ b/model/aro_hcp/v1alpha1/cluster_resource.model
@@ -74,4 +74,9 @@ resource Cluster {
 	locator ControlPlaneUpgradePolicies {
 		target ControlPlaneUpgradePolicies
 	}
+
+	// Reference to cluster resources.
+	locator Resources {
+	    target Resources
+	}
 }
--- a/openapi/aro_hcp/v1alpha1/openapi.json
+++ b/openapi/aro_hcp/v1alpha1/openapi.json
@@ -1705,6 +1705,43 @@
         }
       }
     },
+    "/api/aro_hcp/v1alpha1/clusters/{cluster_id}/resources": {
+      "get": {
+        "description": "Retrieves a list of resources for a cluster in error state",
+        "parameters": [
+          {
+            "name": "cluster_id",
+            "in": "path",
+            "schema": {
+              "type": "string"
+            },
+            "required": true
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Success.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ClusterResources"
+                }
+              }
+            }
+          },
+          "default": {
+            "description": "Error.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/api/aro_hcp/v1alpha1/clusters/{cluster_id}/status": {
       "get": {
         "parameters": [


nimrodshn:
Generated and tested in the following API test https://gitlab.cee.redhat.com/service/aro-hcp-clusters-service/-/merge_requests/368

model/aro_hcp/v1alpha1/resources_resource.model
```diff
@@ -0,0 +1,24 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a collection of resources for a cluster
+resource Resources {
+    // Retrieves a list of resources for a cluster in error state
```
miguelsorianod, line 19: Is the "in error state" part accurate @nimrodshn ? trying to make sure I understand it

model/aro_hcp/v1alpha1/resources_resource.model
```diff
@@ -0,0 +1,24 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Manages a collection of resources for a cluster
+resource Resources {
```
miguelsorianod, line 18: what happens with resources related to nodepools and externalauths? will they also be returned at the cluster level?

nimrodshn: Yes. That is the plan. To avoid N calls for N nodepools, etc.

model/aro_hcp/v1alpha1/resources_resource.model
nimrodshn: Sorry that was copied over from the commercial type. I will fix it asap. Ty.

model/aro_hcp/v1alpha1/resources_resource.model
nimrodshn: Do you have any reservations with that approach? My only slight concern is the http response can grow arbitrarily large. But this mechanism is ment to be a temporary solution and thus isnt designed to be bulletproof.

miguelsorianod, line 18: Do we plan to provide per individual resource get at some point too?

nimrodshn: No. Its just a snapshot of the current resources. The idea is to avoid N requests.

nimrodshn: In resources or types.

nimrodshn: You can also view in more details here https://docs.google.com/document/d/1AMkLLT4I-gOfS_BlH6_32qkSZrHHVGg963pdB54axYM/edit?usp=sharing

model/aro_hcp/v1alpha1/resources_resource.model
nimrodshn: @miguelsorianod Please see #1176

model/aro_hcp/v1alpha1/resources_resource.model
miguelsorianod: > Do you have any reservations with that approach? My only slight concern is the http response can grow arbitrarily large. But this mechanism is ment to be a temporary solution and thus isnt designed to be bulletproof.

Do we know what locking behavior is going to occur and on what resources when this endpoint is exercised? locking can impact the synchronous endpoints in CS as they open DB transactions.

I would expect this to be called relatively frequently, periodically

nimrodshn: > locking can impact the synchronous endpoints in CS as they open DB transactions

When read I dont expect this to take any lock on resources. It will return whatever resources where previously committed during the lifecycle of the resources.

nimrodshn: The underlying SQL is performed on each GET request:
```
 SELECT cluster_id, name, value, creation_timestamp
  FROM cluster_resources
  WHERE cluster_id = $1
```

miguelsorianod: > When read I dont expect this to take any lock on resources. It will return whatever resources where previously committed during the lifecycle of the resources.

Take into account that you do not need to request an explicit lock to have locking. When you do a DB write that also acquires locks implicitly.

Even when you don't do any write you can get locking. The CS API endpoints create DB transactions during the whole lifecycle of the request. For example, if the request does a read on a row only it acquires an Access Share lock, which can be blocked by an ALTER TABLE command for example. The Access Share lock remains not only until the read is finished, but until the whole DB transaction is completed. Which in our case means until the whole API request completes.
If there are other DB interactions in the request other than with the cluster_resources table the same applies, including reads and writes.

PR #1180, opened 2026-07-21
Create ssh public key field

tmstff:
## Description

Implementation of [ARO-26937: Update OCM SDK to include SSH public key field](https://redhat.atlassian.net/browse/ARO-26937) :

As part of the effort to make it possible for ARO HCP customers to connect to a cluster node via ssh, this change adds the fiels nodeSshPublicKey to enable the ARO HCP RP Backend to propagate the key to CS (which can then forward it to HyperShift).

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or tooling change

## Verification

- [x] `make check` passes (model syntax validation)
- [ ] `make lint` passes (indentation enforcement)  -> did not pass, but also on >4.000 locations that were not touched my this PR. Will try to take care of this separately.

## Checklist

- [x] Model files follow the project's DSL conventions (see [README](../README.md#concepts))
- [x] Generated `clientapi/` and `openapi/` are updated if model changed (`make update`)

## New Features

- Added optional SSH public key configuration for cluster nodes to support secure maintenance access.
- Added AWS spot-instance configuration for node pools, including an optional maximum price.
- Added support for configuring the spot-instance termination-handler queue URL.
- Added API, serialization, and schema support for these settings.

## Chores

- Improved linting consistency when scanning model files.


model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -257,4 +257,10 @@ class Cluster {
 	// The AutoNode settings for this cluster.
 	// This is currently only supported for ROSA HCP
 	AutoNode ClusterAutoNode
+
+	// The SSH public key used for secure access to cluster nodes. When set, this key
+	// is distributed to all nodes, enabling SSH access for debugging
+	// and maintenance purposes.
+	// This is currently only supported for ARO HCP
```
miguelsorianod, line 264: > 	// This is currently only supported for ARO HCP

This comment is not needed, as the model is specific to aro-hcp

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -257,4 +257,10 @@ class Cluster {
...
+	NodeSSHPublicKey String
```
miguelsorianod, line 265: Is it updatable? If not, document it in the same way as other fields that are not.

miguelsorianod, line 265: What is the expected format? explain it

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -4920,6 +4920,10 @@
             "description": "Network settings of the cluster.",
             "$ref": "#/components/schemas/Network"
           },
+          "node_ssh_public_key": {
```
miguelsorianod, line 4923: Change to plural: nodes_ssh_public_key

openapi/aro_hcp/v1alpha1/openapi.json
```diff
@@ -4920,6 +4920,10 @@
...
+            "description": "The SSH public key used for secure access to cluster nodes. When set, this key\nis distributed to all nodes, enabling SSH access for debugging\nand maintenance purposes.\nThis is currently only supported for ARO HCP",
```
miguelsorianod, line 4924: cluster nodes -> cluster's node pool's nodes

model/aro_hcp/v1alpha1/cluster_type.model
miguelsorianod: If it's updatable, are both directions supported? document any restrictions

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -257,4 +257,10 @@ class Cluster {
 	// The AutoNode settings for this cluster.
 	// This is currently only supported for ROSA HCP
 	AutoNode ClusterAutoNode
+
+	// The SSH public key used for secure access to cluster nodes. When set, this key
```
miguelsorianod, line 261: same as https://github.com/openshift-online/ocm-api-model/pull/1180/changes#r3621514833

miguelsorianod, line 261: Has the API design been reviewed and approved on ARM side?

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: It's updatable. What do you mean by "both directions" ?

tmstff: valid ssh public key. Can add that.

miguelsorianod: by both directions I mean:
* update from "there's a key set" to "no key set"
* update from "no key set" to "there's a key set"

miguelsorianod: > It's updatable. 

Do we have an understanding on how that would be achieved (at RP level, CS level, ...) and what would be the impact (recreation of nodepools vs not, etc...) ? this is a cluster-level attribute but it controls node pool related configuration. This would include figuring out how to perform update state calculation and so on.

Do we know whether Hypershift supports updating it? if so, what are the consequences of doing so.

model/aro_hcp/v1alpha1/cluster_type.model
miguelsorianod: The ARM side PR is https://github.com/Azure/ARO-HCP/pull/6157, which is currently in draft. I left some comments there too

openapi/aro_hcp/v1alpha1/openapi.json
tmstff: discussed it with Ben: we want to leave it as node_ssh_public_keys (added an s at the end because it can be many) because it's clear enough this way already.

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: My understanding is, that you can update the value (also with an empty value) and this will trigger the recreation of all nodepools. This is backed by the following documentation:

> 	// When sshKey is set, the controllers will generate a machineConfig with the sshAuthorizedKeys https://coreos.github.io/ignition/configuration-v3_2/ populated with this value.
>	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
>	// Changing this value will trigger a rollout for all existing NodePools in the cluster.

I have verified that you can change keys, but not yet that providing an empty list will reset. Will try that out.

Would you want this behavior documented here?

tmstff: > how that would be achieved (at RP level, CS level, ...)

RP passes the updated value to CS, CS updates the secret containing the keys - and also the CR (e.g. label with hash from the keys) to trigger reconcile.

tmstff: @miguelsorianod let me know what you think

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -265,4 +265,10 @@ class Cluster {
 	// The AutoNode settings for this cluster.
 	// This is currently only supported for ROSA HCP
 	AutoNode ClusterAutoNode
+
+	// SSH public keys used for secure access to the cluster's node pool's nodes. When set, keys
+	// are distributed to all nodes, enabling SSH access for debugging and maintenance purposes.
+	// Each entry must contain a single SSH public key that would be acceptable for
+	// an authorized key file (`.ssh/authorized_keys`).
+	NodeSSHPublicKeys []String
```
miguelsorianod, line 278: During a CS PATCH, how will we express:
* removing a key from the list of **existing** keys?
* adding a key to a list of **existing** keys?

Take into account CS possible limitations, PATCH semantics and limitations on arrays and so on

model/aro_hcp/v1alpha1/cluster_type.model
miguelsorianod: We need to have an accurate understanding and document it appropriately in both the CS and RP APIs

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -265,4 +265,10 @@ class Cluster {
 	// The AutoNode settings for this cluster.
 	// This is currently only supported for ROSA HCP
 	AutoNode ClusterAutoNode
+
+	// SSH public keys used for secure access to the cluster's node pool's nodes. When set, keys
+	// are distributed to all nodes, enabling SSH access for debugging and maintenance purposes.
+	// Each entry must contain a single SSH public key that would be acceptable for
+	// an authorized key file (`.ssh/authorized_keys`).
```
miguelsorianod: > Each entry must contain a single SSH public key that would be acceptable for an authorized key file (`.ssh/authorized_keys`)

Does that mean **all** possible formats supported by .ssh/authorized_keys are supported? this is signaling that. If not, this is not correct and we need to be more precise

tmstff: yes, we do. Essentially we just concatenate the ssh keys with \n and everything gets dumped into the authorized_keys file.

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: You always replace the whole array - that's standard with http PUT and PATCH (in case the field is provided), no?

so if you want to delete one key, you remove it from the array you sent last time and re-apply it again.

I assumed this to be default behavior. But I can document it anyways to avoid confusion.

Could you please elaborate on what you mean with "CS possible limitations" and "limitations on arrays and so on"? @miguelsorianod

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: The description now matches the desired behavior. Same changes will be needed to implement that, but the hypershift team and I are going to take care of it.

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: Updated the documentation.

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -265,4 +265,15 @@ class Cluster {
 	// The AutoNode settings for this cluster.
 	// This is currently only supported for ROSA HCP
 	AutoNode ClusterAutoNode
+
+	// One or more SSH public keys used for secure access to worker nodes. When set,
+	// these keys are distributed to all nodes, enabling SSH access for debugging
+	// and maintenance purposes. If changed, all keys previously set will be replaced.
```
machi1990, line 271: I suppose changing the keys causes nodes to be rebooted, correct? If so, let's document it.
See the corresponding description from Hypershift https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L702

> // sshKey is a local reference to a Secret that must have a "id_rsa.pub" key whose content must be the public part of 1..N SSH keys.
	// If the reference is set but none of the above requirements are met, the HostedCluster will enter a degraded state.
	// TODO(alberto): Signal this in a condition.
	// When sshKey is set, the controllers will generate a machineConfig with the sshAuthorizedKeys https://coreos.github.io/ignition/configuration-v3_2/ populated with this value.
	// This MachineConfig will be part of every payload generated by the controllers for any NodePool of the HostedCluster.
	// Changing this value will trigger a rollout for all existing NodePools in the cluster.

model/aro_hcp/v1alpha1/cluster_type.model
```diff
@@ -265,4 +265,15 @@ class Cluster {
...
+	//
+	// WARNING: Setting or changing this field currently causes the recreation of all worker nodes.
+	// This behavior is subject to change in a future OpenShift release.
+	//
+	// SSH public key format: <algorithm> <base64-encoded-key> [comment]
+	// Key constraints: algorithm + space + base64 blob required; comment optional; no newlines within key.
```
nimrodshn, line 277: Are these constraints validated in CS or RP? Can you please point to the implementation PR in CS? 

Best effort: I would also write a small e2e with a locally generated sdk to ensure this change works correctly.

tmstff: They are validated in RP:
- https://github.com/Azure/ARO-HCP/pull/6157/changes/BASE..a398fcbe50f49ae17e9d0a16ee83ca56a932a01c#diff-37e8daa623bf0808fb4a6d552f215f211d40031e8c0add959a95d38345e8e1c1
- https://github.com/Azure/ARO-HCP/pull/6157/changes/BASE..a398fcbe50f49ae17e9d0a16ee83ca56a932a01c#diff-c0b43921e58d549340662b0ebbcd1488f4ef1bf3702dcb9b7ac165368c0d7457

There is no implementation PR in CS yet, as I was waiting for this PR to be merged, to get the changes into the SDK, to then use it in CS. Is there a better approach? Sorry, new to CS development ...

Should I replicate the validation in CS, or is RP sufficient?

concerning e2e tests: I've got an e2e test here: https://github.com/Azure/ARO-HCP/pull/6157/changes/BASE..a398fcbe50f49ae17e9d0a16ee83ca56a932a01c#diff-b4bd13a00cc9e5d93c2df36d584530ece6d548b789660688821320e0db07ad66

It's currently not including CS, but my plan is to test the whole chain (once the implementation has progressed further): set keys via arm, and then connect to worker node via ssh (will need some networking setup, but should be doable). Or are there other e2e test you were thinking about @nimrodshn ?

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: @machi1990 is line 273 sufficient? ( WARNING: Setting or changing this field currently causes the recreation of all worker nodes.)

machi1990: Yes, that's correct, I totally missed it. Sorry

model/aro_hcp/v1alpha1/cluster_type.model
nimrodshn: I ment e2e and implementation in CS? is this implemented in CS?

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: No problem :-)

model/aro_hcp/v1alpha1/cluster_type.model
tmstff: No, this is what I meant by 

> There is no implementation PR in CS yet, as I was waiting for this PR to be merged, to get the changes into the SDK, to then use it in CS. Is there a better approach? Sorry, new to CS development ...

Open to better ideas @nimrodshn :-)

PR #1186, opened 2026-08-10
feat: aro-hcp: allow providing CS Provision Shard ID during cluster creation

miguelsorianod:
The placement decision is going to be outsourced outside of CS. This
allows CS receiving the provision shard ID directly instead of
leveraging the CS Provision Shard pinning mechanism based on the
properties map which is fragile.

Additionally, we remove the ProvisionShardID attribute in the Cluster type that existed. It was not being used nor referenced in inbound nor outbound mappers.

## Summary by CodeRabbit

* **New Features**
  * Clusters can now be assigned using a provision shard ID.
  * When no ID is provided, the system automatically selects a suitable shard.
  * Updated cluster APIs and schemas expose the new `provision_shard_id` field.

* **Breaking Changes**
  * The previous provision shard object and related accessors have been replaced by the provision shard ID string.
  * Providing an ID bypasses automatic capacity and shard-state checks.

### DIFF
--- a/clientapi/arohcp/v1alpha1/cluster_builder.go
+++ b/clientapi/arohcp/v1alpha1/cluster_builder.go
@@ -112,7 +112,7 @@ type ClusterBuilder struct {
 	openshiftVersion                  string
 	product                           *ProductBuilder
 	properties                        map[string]string
-	provisionShard                    *ProvisionShardBuilder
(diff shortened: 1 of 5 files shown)


machi1990 (commented):
/lgtm

